PR #165, opened 2018-06-16
Remove min connection count for block header syncing

doraemon0:


yilunzhang:
We can remove it for now, but I think having a minimal neighbor count >= 3 is reasonable, otherwise the voting doesn't make sense

doraemon0:
@yilunzhang Will handle it in consensus layer. 
If the node count started by tester is not enough for 4, the nkn program will be blocked always.

PR #172, opened 2018-06-24
Add support for multiple clients sharing the same NKN address

yilunzhang:
Add support for multiple clients sharing the same NKN address
All clients sharing the address will receive msg

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -16,6 +16,7 @@ import (
 	"github.com/nknorg/nkn/util/log"
 	"github.com/nknorg/nkn/vault"
 	"github.com/nknorg/nkn/websocket"
+	"github.com/nknorg/nkn/websocket/session"
 )
 
 type RelayService struct {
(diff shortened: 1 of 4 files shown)


relay/relay.go
```diff
@@ -163,8 +167,9 @@ func (rs *RelayService) HandleMsg(packet *message.RelayPacket) error {
 	if nextHop == nil {
 		clients := websocket.GetServer().GetClientsById(destID)
 		if clients == nil {
-			// TODO: handle client not exists
-			return errors.New("Client Not Exists: " + hex.EncodeToString(destID))
+			log.Info("Client Not Online:", hex.EncodeToString(destID))
+			rs.addRelayPacketToBuffer(destID, packet)
```
dreamfly281: Any space limitation for  buffer the packets

yilunzhang: Not yet. Currently it's unlimited.

PR #174, opened 2018-06-25
[Changed] Add/init webSockPort & httpJSONPort in net/node

gdmmx:
By the way, fix httpInfoPort from int to uint16


yilunzhang:
Looks good to me

PR #177, opened 2018-06-26
Use the same interface with httpjson in websocket and delete httprestful 

NoelBright:
1. move rpc/ to api/ and move websocket/ into api/  
2. Use the same interface with httpjson in websocket  
3. make websocket independence on httprestful  
4. delete httprestful module


NoelBright:
@yilunzhang  done.

PR #179, opened 2018-06-27
block syncing feature

doraemon0:
block syncing contains two steps:
a) sync blocks from neighbors
b) cache proposed block when syncing block. after block syncing finished, verify and persist cached block to DB.

flow:
1. listen to BlockFlooding message, then local node knows which block is
stop block hash to sync.
2. request headers and blocks from a random neighbor, at same time, cache the
proposed blocks in order.
3. after block syncing finished, get blocks from cache, verify them
in order then persist them.
4. when all blocks(synced and cached) are persisted, start to do consensus.

note:
1. Don't accept any transactions to transaction pool if block syncing is not finished.
2. Orphan block cache is no longer required. Move and modify it to be a cache for syncing block.

TODO:
1. choose a >=50% neighbor to sync block
2. vote counting for cached proposed block rather than save directly.


doraemon0:
@dreamfly281 yes. the error is use stop hash as start hash.

There's another error when both start hash and stop hash are not empty in request message. But the code path is never being called.

PR #184, opened 2018-06-29
[Changed] Adjust args flags of nknd. Indicate SeedList in os.Args

gdmmx:
1. Indicate create Mode via -c flag
2. Support indicate SeedList override config.json via "-seed str" flag
3. Adjust README & launch script for consistent with args changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,13 +127,13 @@ conflict in ports will be resolved automatically.
 Start bootstrap node by creating a network
 
 ```shell
-$ ./nknd -test create
+$ ./nknd -c
 ```
 
 Start other nodes by joining the network
 
 ```shell
-$ ./nknd -test join
+$ ./nknd -seed $RemoteNodeIP:$HttpJsonPort
 ```
 
 When the network contains enough nodes (usually 8+), stop the node that created
@@ -161,13 +161,13 @@ $ docker run -v $PWD:/nkn -it nkn nknc wallet -c
 Start bootstrap node by creating a network
 
 ```shell
-$ docker run -p 30000-30003:30000-30003 -v $PWD:/nkn -it nkn nknd -test create
+$ docker run -p 30000-30003:30000-30003 -v $PWD:/nkn -it nkn nknd -c
 ```
 
 Start other nodes by joining the network
 
 ```shell
-$ docker run -p 30000-30003:30000-30003 -v $PWD:/nkn -it nkn nknd -test join
+$ docker run -p 30000-30003:30000-30003 -v $PWD:/nkn -it nkn nknd -seed $RemoteNodeIP:$HttpJsonPort
 ```
 
 ## Contributing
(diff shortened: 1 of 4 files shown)


nknd.go
```diff
@@ -157,8 +162,33 @@ func nknMain() error {
...
+			Name:        "seed",
+			Usage:       "Seed List to join",
+			Value:       "localhost:30003",
+			Destination: &seedStr,
+		},
+		cli.StringFlag{
+			Name:        "passwd, p",
+			Usage:       "Password of Your wallet private Key",
+			Value:       "",
+			Destination: &seedStr,
```
yilunzhang: Is this a typo?

util/password/password.go
```diff
@@ -47,16 +46,11 @@ func GetConfirmedPassword() ([]byte, error) {
...
-	flag.Parse()
-	if pstr == "" {
-		passwd, err = GetPassword()
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		passwd = []byte(pstr)
+
+	// Don't support input passwd in os.Args as plain text
```
yilunzhang: The CLI has a -p params now, which seems to be conflicting with this comment

gdmmx: cmdline  -p params is NOT safe. The password would be let out via bash_history.
I suggest cancel the -p flag, only input password via GetPassword() which without echo.

yilunzhang: Agree. Or we can keep it but only use it for testnet cluster (not put it in help)

gdmmx:
bu@yilunzhang I supplemented a fix f59cdcd for comply with your comments.

README.md
```diff
@@ -127,13 +127,13 @@ conflict in ports will be resolved automatically.
...
 ```shell
-$ ./nknd -test create
+$ ./nknd -c
 ```
 
 Start other nodes by joining the network
 
 ```shell
-$ ./nknd -test join
+$ ./nknd -seed $RemoteNodeIP:$HttpJsonPort
```
dreamfly281: Is there any commend for normal user joining the network without set any parameter?

gdmmx: @dreamfly281 SeedList is a MUST parameter which the entry points during you join a ring from. It can not be omitted. Previous cmdline "-test join" just implied SeedList setting via config.json.
New cmdline "--seed" just provide another approach to override SeedList in config.json. If --seed indicate an empty string "", the setting in config.json is still efficient.
```
if len(seedStr) > 0 {
    // Just override config.json when --seed length > 0
    config.Parameters.SeedList = strings.Split(seedStr, ",")
}
```
Currently, the default value of --seed is "localhost:30003"

yilunzhang: We can have another line like:
To join the network using the seed list defined in config.json
./nknd

gdmmx: @dreamfly281 @yilunzhang I supplemented a fix 107457a for used config.json setting as default when no --seed params provided.

PR #185, opened 2018-07-02
improvements for block syncing

doraemon0:
1. non-proposer node waits for first block flooding message to
start syncing.
2. if proposer node is lower than neighbors, sync the longest
chain first.
3. wait for syncing finished then start consensus
4. optimizing log


doraemon0:
Another commit about consensus restart will be append to this PR later.

PR #187, opened 2018-07-06
change block proposer when current block proposer not available

doraemon0:
1. exit progress directly first when node has different consensus state especially for keepalive timeout then reconnect.
2. if current block proposer not available then change another one based on timestamp and history.


dreamfly281:
This patch is under testing,  could  be merged after the testing running at 24 hours or even longer to make sure it works as expectation.

PR #190, opened 2018-07-10
[BugFix] IsAddrInNbrList's arg $addr was overrided by local var

gdmmx:
strings.Compare(add+":"+Itoa(port),  addr) will never == 0.


doraemon0:
Good catch!

PR #192, opened 2018-07-10
Remove the duplicate code and avoid warning when create ring

dreamfly281:
It doesn't need to connect the seed node to get the blockheight
and chord DHT port when create the initial node.

Signed-off-by: Yanbo Li <dreamfly281@gmail.com>


gdmmx:
Looks good to me.

gdmmx:
Looks good to me.

PR #195, opened 2018-07-12
 Voting for cached blocks when sync block

doraemon0:
Detecting neighbor sync state by using ping and pong messages.
Neighbor node which in syncing mode will be ignored while local node
syncing block.

Listen to neighbor's proposal and mind changing messages when
sync block to determine if cached block which proposed by
block proposer is acceptable. The stop hash of block sycing will be
decided by the first cached block.

### DIFF
--- a/consensus/ising/proposerservice.go
+++ b/consensus/ising/proposerservice.go
@@ -81,6 +81,9 @@ func NewProposerService(account *vault.Account, node protocol.Noder) *ProposerSe
 	if !service.timeout.Stop() {
 		<-service.timeout.C
 	}
+	if !service.proposerChangeTimer.Stop() {
+		<-service.proposerChangeTimer.C
+	}
 
 	return service
 }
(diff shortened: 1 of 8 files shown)


consensus/ising/proposerservice.go
```diff
@@ -755,15 +762,27 @@ func (ps *ProposerService) SetOrChangeMind(votingType voting.VotingContentType,
...
 	if ps.localNode.GetSyncState() != protocol.PersistFinished {
+		err := ps.syncCache.AddVoteForBlock(hash, height, nodeID)
+		if err != nil {
+			return
+		}
+		block, err := ps.syncCache.GetBlockFromSyncCache(height)
+		if err != nil {
+			return
+		}
+		ps.localNode.SetSyncStopHash(block.Header.PrevBlockHash, block.Header.Height-1)
```
yilunzhang: One potential security concern for this approach is that malicious node can continuously send wrong block to nodes in sync mode to prevent it from finish syncing

consensus/ising/synccache.go
```diff
@@ -1,103 +1,212 @@
...
+		// cached block height [min height, max height)
 		sc.minHeight = blockHeight
-		sc.maxHeight = blockHeight
+		sc.maxHeight = blockHeight + 1
 	} else {
-		if blockHeight != sc.maxHeight+1 {
+		if blockHeight != sc.maxHeight {
 			return errors.New("adding block which height is invalid")
 		}
 		sc.maxHeight++
```
yilunzhang: What if node receives multiple blocks of the same height?

doraemon0: All received blocks with same height will be cached. Only the block which has >50% votes will be persisted.

consensus/ising/proposerservice.go
doraemon0: There's a flag in SetSyncStopHash function when setting stop hash. If there's a valid stop hash set, the later will not be set again.

consensus/ising/synccache.go
yilunzhang: But is "sc.maxHeight++" called every time when the node receives a block and try to add it to sync cache?

doraemon0: Good catch, it's a bug, let me fix it.

consensus/ising/proposerservice.go
doraemon0: The first one is voted by neighbors. If most neighbors are evil, let it go.

PR #199, opened 2018-07-16
changes for block proposer verification

doraemon0:
1. Exclusive signature chain transactions from block when sigchain height
is current height + 3. Making sure the sigchain transaction could be packaged in previous block.

2. Modularize block mining and add block signing.

3. cache block proposer info in ProposerCache, cache more things such as winning info(hash and hash type).

4. add winning info(hash and hash type) to block header for verifying next block.

### DIFF
--- a/consensus/ising/proposerservice.go
+++ b/consensus/ising/proposerservice.go
@@ -2,19 +2,16 @@ package ising
 
 import (
 	"fmt"
-	"math/rand"
 	"os"
 	"sync"
 	"time"
 
 	. "github.com/nknorg/nkn/common"
 	"github.com/nknorg/nkn/consensus/ising/voting"
-	"github.com/nknorg/nkn/core/contract/program"
 	"github.com/nknorg/nkn/core/ledger"
 	"github.com/nknorg/nkn/core/transaction"
 	"github.com/nknorg/nkn/core/transaction/payload"
 	"github.com/nknorg/nkn/crypto"
-	"github.com/nknorg/nkn/crypto/util"
 	"github.com/nknorg/nkn/events"
 	"github.com/nknorg/nkn/net/message"
 	"github.com/nknorg/nkn/net/protocol"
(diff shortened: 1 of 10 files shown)


consensus/ising/voting/sigchain.go
```diff
@@ -95,7 +95,7 @@ func (scv *SigChainVoting) SetVotingHeight(height uint32) {
 }
 
 func (scv *SigChainVoting) GetVotingHeight() uint32 {
-	return ledger.DefaultLedger.Store.GetHeight() + 2
+	return ledger.DefaultLedger.Store.GetHeight() + 3
```
dreamfly281: No magic number, use const name or explain why 2&3 be used here

core/ledger/jsontype.go
```diff
@@ -14,6 +14,7 @@ type HeaderInfo struct {
 	ConsensusData    uint64              `json:"consensusData"`
 	NextBookKeeper   string              `json:"nextBookKeeper"`
 	Signer           string              `json:"signer"`
+	Signature        string              `json:"signature"`
```
dreamfly281: This commit looks good from my side

consensus/ising/proposerservice.go
```diff
@@ -133,22 +129,9 @@ func (ps *ProposerService) ConsensusRoutine(vType voting.VotingContentType) {
...
-			txn := content.(*transaction.Transaction)
-			payload := txn.Payload.(*payload.Commit)
-			sigchain := &por.SigChain{}
-			proto.Unmarshal(payload.SigChain, sigchain)
-			// TODO: get a determinate public key on signature chain
-			pbk, err := sigchain.GetLedgerNodePubkey()
-			if err != nil {
-				log.Warn("Get last public key error", err)
-				return
+			if txn, ok := content.(*transaction.Transaction); ok {
```
dreamfly281: "*transaction.Transaction" ?

doraemon0: Transaction structure in transaction package. We could make this more logical when re-architect.

consensus/ising/voting/sigchain.go
doraemon0: Fixed.

PR #200, opened 2018-07-16
Add mining field to sigchain

doraemon0:
1. Using mining field to indicate whether the singer will take apart in
mining. Singer only transmit packages rather than propose block
when node in block syncing state. Set it to false for client by default.

2. The sigchain will not be cached in sigchain cache(PorServer) when all mining fields are false.

3. calculate determinate position for block proposer on signature chain. 


doraemon0:
@dreamfly281 The block proposer is selected from signature chain. Sometimes node in signature chain won't mine or can't mine, for example, in syncing mode the node can't mine due to no latest block. So node could set the "mining" flag to false when he will not mine. If the flag set to false for signature chain element, the node will  not be selected.

PR #201, opened 2018-07-18
[feature] add authorization function in jsonRPC

NoelBright:
[api/httpjson]
1. add interface Authorization for authentication.
2. add structure UserPassword to implement Authorization.

### DIFF
--- a/api/httpjson/auth.go
+++ b/api/httpjson/auth.go
@@ -1,0 +1,66 @@
+package httpjson
+
+import (
+	"crypto/sha256"
+	"crypto/subtle"
+	"net/http"
+	"strings"
+)
+
+type RPCAuthType byte
+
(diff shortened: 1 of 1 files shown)


api/httpjson/auth.go
```diff
@@ -0,0 +1,66 @@
...
+	"strings"
+)
+
+type RPCAuthType byte
+
+const (
+	UsernamePassword RPCAuthType = 0x01
+)
+
+type Autherization interface {
```
RealJohnSmith: Looks like typo. It should be 'Authorization'

NoelBright: thanks!

NoelBright:
@dreamfly281 When Setting miner address or toggle mining.

PR #210, opened 2018-07-25
block proposer validity verification

doraemon0:

The block proposer validity should be verified by following condition:
a) First 0,1,2,3,4, blocks should be signed by GenesisBlockProposer.
b) Blocks should be signed by a node on sigchain transaction which
in previous block when regular run.
c) When proposer changed due to timeout, the block should be signed
by signer of historical block.
d) If no any sigchain transactions, the block should be signed by
previous block signer.

### Checklist

### Extra information

### DIFF
--- a/consensus/ising/proposerservice.go
+++ b/consensus/ising/proposerservice.go
@@ -14,17 +14,16 @@ import (
 	"github.com/nknorg/nkn/events"
 	"github.com/nknorg/nkn/net/message"
 	"github.com/nknorg/nkn/net/protocol"
+	"github.com/nknorg/nkn/util/config"
 	"github.com/nknorg/nkn/util/log"
 	"github.com/nknorg/nkn/vault"
 )
 
 const (
 	TxnAmountToBePackaged      = 1024
-	ConsensusTime              = 10 * time.Second
 	WaitingForFloodingFinished = time.Second * 1
 	WaitingForVotingFinished   = time.Second * 5
 	TimeoutTolerance           = time.Second * 2
-	ProposerChangeTime         = time.Minute
 )
 
 type ProposerService struct {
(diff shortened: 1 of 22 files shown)


consensus/ising/proposerservice.go
```diff
@@ -14,17 +14,16 @@ import (
...
 	"github.com/nknorg/nkn/net/protocol"
+	"github.com/nknorg/nkn/util/config"
 	"github.com/nknorg/nkn/util/log"
 	"github.com/nknorg/nkn/vault"
 )
 
 const (
 	TxnAmountToBePackaged      = 1024
-	ConsensusTime              = 10 * time.Second
 	WaitingForFloodingFinished = time.Second * 1
```
yilunzhang: If we move ConsensusTime to config, probably we should move all these constants to config also?

core/ledger/header.go
```diff
@@ -17,9 +17,15 @@ import (
...
-	WinningBlockHash WinningHashType = 2
+	// The proof of Block proposer validity should exists in previous Block header.
+	// GenesisHash means next Block proposer is GenesisBlockProposer.
+	GenesisHash WinningHashType = 0
+	// WinningTxnHash means next Block proposer is a node on signature chain.
+	WinningTxnHash WinningHashType = 1
+	// WinningNilHash means next Block proposer is the signer of last Block.
+	WinningNilHash WinningHashType = 2
+	// WinningBlockHash means next Block proposer is signer of historical Block.
+	WinningBlockHash WinningHashType = 3
```
yilunzhang: Is there any reason why we want to separate WinningNilHash from WinningBlockHash?

yilunzhang:
Made a few comments. The rest looks good to me

consensus/ising/proposerservice.go
doraemon0: Moving the important time ConsensusTime and ProposerChangeTime to config since other modules may depend on these parameter, e.g.:block proposer verification when timeouts.
Other constants can continue to be placed in the consensus module.

core/ledger/header.go
doraemon0: When no signature chain in system, the last block singer will produce and sign next block. WinningNilHash is only used for verifying block singer, the singer must be signer of last block when no signature chain.

When block singer(could be last block signer or consensused by signature chain) is not available, we do block proposer changing according to timeout. WinningBlockHash is only used for tracking the block proposer which is chosen by timeout from historical block.

PR #215, opened 2018-08-02
broadcast transaction

doraemon0:
broadcast transation through relay to neighbors

### Checklist


net/node/node.go
```diff
@@ -375,6 +376,22 @@ func (node *node) Xmit(message interface{}) error {
 	return nil
 }
 
+func (node *node) Relay(from Noder, txn *transaction.Transaction) error {
```
yilunzhang: Should we change the name Relay here? It's different from our Proof of Relay and may cause some confusions. What about BroadcastTransaction?

doraemon0: Good idea, let me fix it.

net/node/node.go
```diff
@@ -375,6 +377,22 @@ func (node *node) Xmit(message interface{}) error {
...
+	buffer, err := NewTxn(txn)
+	if err != nil {
+		log.Error("Error New Tx message: ", err)
+		return err
+	}
+	node.txnCnt++
+	for _, n := range node.GetSyncFinishedNeighbors() {
+		if n.GetRelay() && n.GetID() != from.GetID() {
+			n.Tx(buffer)
+		}
```
dreamfly281: What will happened if the node connection lost after get the syncfinishedneighbors list?

PR #228, opened 2018-08-20
dynamic voting in consensus and block syncing

doraemon0:
dynamic voting in consensus and block syncing

### DIFF
--- a/consensus/ising/voting/blockvotingpool.go
+++ b/consensus/ising/voting/blockvotingpool.go
@@ -5,24 +5,23 @@ import (
 	"sync"
 
 	. "github.com/nknorg/nkn/common"
-	"github.com/nknorg/nkn/core/ledger"
 	"github.com/nknorg/nkn/util/log"
 )
 
 type BlockVotingPool struct {
 	sync.RWMutex
-	sendPool    map[uint32]map[uint64]Uint256 // record voting history
-	receivePool map[uint32]map[uint64]Uint256 // received votes from neighbor
-	mind        map[uint32]Uint256            // current idea
+	sendPool    map[uint32]map[uint64]Uint256   // record voting history
+	receivePool map[uint32]map[uint64]*VoteInfo // received votes from neighbor
+	mind        map[uint32]Uint256              // current idea
 	totalWeight int
 }
 
-func NewBlockVotingPool(totalWeight int) *BlockVotingPool {
+func NewBlockVotingPool() *BlockVotingPool {
 	return &BlockVotingPool{
 		sendPool:    make(map[uint32]map[uint64]Uint256),
-		receivePool: make(map[uint32]map[uint64]Uint256),
+		receivePool: make(map[uint32]map[uint64]*VoteInfo),
 		mind:        make(map[uint32]Uint256),
-		totalWeight: totalWeight,
+		totalWeight: 0,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


consensus/ising/voting/sigchainvotingpool.go
```diff
@@ -80,34 +71,65 @@ func (scvp *SigChainVotingPool) AddToSendPool(height uint32, nid uint64, hash Ui
 	}
 }
 
-func (scvp *SigChainVotingPool) VoteCounting(height uint32) (*Uint256, error) {
-	scvp.RLock()
-	defer scvp.RUnlock()
+func (scvp *SigChainVotingPool) AddToReceivePool(height uint32, nodeID uint64, weight int, hash Uint256) {
```
yilunzhang: Why we need two methods addToReceivePool and AddToReceivePool instead of just AddToReceivePool?

consensus/ising/voting/sigchainvotingpool.go
```diff
@@ -80,34 +71,65 @@ func (scvp *SigChainVotingPool) AddToSendPool(height uint32, nid uint64, hash Ui
...
+			m[vHash] += vWeight
 		} else {
-			m[hash] = weight
+			m[vHash] = vWeight
 		}
 	}
+
 	// returns hash which got >=50% votes
 	for hash, count := range m {
 		if 2*count >= scvp.totalWeight {
```
yilunzhang: Same as above.

consensus/ising/voting/blockvotingpool.go
```diff
@@ -81,35 +71,65 @@ func (bvp *BlockVotingPool) AddToSendPool(height uint32, nid uint64, hash Uint25
 	}
 }
 
-func (bvp *BlockVotingPool) VoteCounting(height uint32) (*Uint256, error) {
-	bvp.RLock()
-	defer bvp.RUnlock()
+func (bvp *BlockVotingPool) AddToReceivePool(height uint32, nodeID uint64, weight int, hash Uint256) {
```
yilunzhang: Same question here.

consensus/ising/voting/sigchainvotingpool.go
doraemon0: Both AddToReceivePool and AddVoteThenCounting need to add vote to pool. Function addToReceivePool is common function for them.

consensus/ising/voting/blockvotingpool.go
```diff
@@ -81,35 +71,65 @@ func (bvp *BlockVotingPool) AddToSendPool(height uint32, nid uint64, hash Uint25
...
 	return nil, errors.New("invalid votes")
 }
 
-func (bvp *BlockVotingPool) Reset(totalWeight int) {
-	bvp.totalWeight = totalWeight
+func (bvp *BlockVotingPool) Reset() {
+	bvp.totalWeight = 0
+}
+
+func (bvp *BlockVotingPool) addToReceivePool(height uint32, nodeID uint64, weight int, hash Uint256) {
```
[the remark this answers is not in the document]
doraemon0: addToReceivePool is common function for AddToReceivePool and AddVoteThenCounting.

consensus/ising/voting/blockvotingpool.go
```diff
@@ -81,35 +71,65 @@ func (bvp *BlockVotingPool) AddToSendPool(height uint32, nid uint64, hash Uint25
...
+			m[vHash] += vWeight
 		} else {
-			m[hash] = weight
+			m[vHash] = vWeight
 		}
 	}
+
 	// returns hash which got >=50% votes
 	for hash, count := range m {
 		if 2*count >= bvp.totalWeight {
```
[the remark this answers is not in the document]
doraemon0: After talk about this question with @gdmmx offline, I'd like to use > in all places. How do you think about it？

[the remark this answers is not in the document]
doraemon0: Which means following votes is enough for (1/1, 2/2, 2/3, 3/4, 3/5...)  finalizing a block.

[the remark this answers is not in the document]
yilunzhang: Yes > makes sense to me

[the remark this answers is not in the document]
yilunzhang: Then we need to make changes to those ">=" like this one

doraemon0:
Change to >50% and fix the conflicts.

consensus/ising/voting/sigchainvotingpool.go
dreamfly281: so could we rename them to avoid confuse?

doraemon0:
Testing on cluster2, in progress.

doraemon0:
Add a new patch for this PR.
Found in testing:
When lots of new node join network together, the new joined node may be surrounded by nodes which all in syncing mode(will not send vote), so the block cached in SyncCache may don't have enough votes. Add a consensus height for SyncCache to indicate the first height which got enough votes from neighbors. Will persist cached blocks from consensus height when block syncing finished.

consensus/ising/voting/sigchainvotingpool.go
doraemon0: Maybe we could make this to be common coding style.
For example, 
function InitTest() can call initTest()，
InitTest() is for public use and initTest() for internal use only.

yilunzhang: I think we can do it

yilunzhang:
Looks good to me. We can merge if testing results are positive

PR #230, opened 2018-08-20
add length check for invalid encoded public key

doraemon0:
Add length check for invalid encoded public key.
For compressed length, it should be 33 bytes, 0x02 or 0x03 || X
For uncompressed length, it should be 65 bytes, 0x04 || X || Y


gdmmx:
LGTM

PR #239, opened 2018-08-27
code changes for issues found during testing

doraemon0:
1. don't handle block persistence when in syncing mode
2. malloc memory when vote comes before block
3. fix proposer index out of bounds when timeout is too long
4. sync to block which got enough votes rather than previous block
5. print more logs for block proposer verification
6. avoid generate huge log and limit goroutine count

### Extra information
Already tested on cluster 3 with about 500 nodes.


yilunzhang:
Looks good to me

PR #243, opened 2018-08-28
Fix bug that cause wrong network topology

yilunzhang:
1. If connection is closed when a remote node is in connecting list, the remote node should be removed from connecting list, otherwise the remote node will never be connected again.
2. When comparing chord id, shorter length id should always be less than longer id (effectively add zero padding in front of chord id that is shorter than hashbit).

### Checklist

### Extra information

### DIFF
--- a/net/node/node.go
+++ b/net/node/node.go
@@ -118,13 +118,12 @@ func (node *node) SetAddrInConnectingList(addr string) (added bool) {
 func (node *node) RemoveAddrInConnectingList(addr string) {
 	node.ConnectingNodes.Lock()
 	defer node.ConnectingNodes.Unlock()
-	addrs := []string{}
 	for i, a := range node.ConnectingAddrs {
 		if a == addr {
-			addrs = append(node.ConnectingAddrs[:i], node.ConnectingAddrs[i+1:]...)
+			node.ConnectingAddrs = append(node.ConnectingAddrs[:i], node.ConnectingAddrs[i+1:]...)
+			return
 		}
 	}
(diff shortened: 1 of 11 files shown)


net/node/node.go
```diff
@@ -147,6 +146,7 @@ func NewNode() *node {
 	n := node{
 		state: INIT,
 	}
+	n.time = time.Now()
```
gdmmx: Just a suggestion：
How about below？
```
	n := node{
 		state: INIT,
                time:  time.Now()，
 	}
```

yilunzhang: It didn't work, because time is a property of node.link but not node.

net/node/node.go
```diff
@@ -658,8 +659,10 @@ func (node *node) ConnectNeighbors() {
 		if !node.IsChordAddrInNeighbors(chordNbr.Id) {
 			chordNbrAddr, err := chordNbr.NodeAddr()
 			if err != nil {
+				log.Error("Get chord neighbor node addr error:", err)
```
gdmmx: There was logged err in chordNbr.NodeAddr() already.  [NodeAddr()](https://github.com/nknorg/nkn/blob/master/net/chord/vnode.go#L26)
Why log it again at here?

net/node/node.go
gdmmx: My mistake. Sorry about that.

net/node/node.go
yilunzhang: You are right, removing it

PR #259, opened 2018-09-03
limit goroutine number for handling message

doraemon0:
goroutine will be created to handle message when message comes. If lots of message reached together, lots of goroutines will be created. It too heavy for system. Limiting the goroutine number in this patch. 

- [x] Enhancement

### DIFF
--- a/net/node/chanqueue.go
+++ b/net/node/chanqueue.go
@@ -1,0 +1,10 @@
+package node
+
+type ChanQueue chan struct{}
+
+func MakeChanQueue(n int) ChanQueue {
+	return make(chan struct{}, n)
+}
+
+func (s ChanQueue) acquire() { s <- struct{}{} }
+func (s ChanQueue) release() { <-s }
--- a/net/node/node.go
+++ b/net/node/node.go
@@ -28,6 +28,7 @@ import (
 
 const (
 	MaxSyncHeaderReq     = 2                // max concurrent sync header request count
+	MaxMsgChanNum        = 2048             // max goroutine num for message handler
 	ConnectionMaxBackoff = 4000             // back off for retry
 	MaxRetryCount        = 3                // max retry count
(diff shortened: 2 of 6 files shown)


net/node/node.go
```diff
@@ -28,6 +28,7 @@ import (
 
 const (
 	MaxSyncHeaderReq     = 2                // max concurrent sync header request count
+	MaxMsgChanNum        = 2048             // max goroutine num for message handler
```
yilunzhang: Is 2048 too small?

net/node/chanqueue.go
```diff
@@ -0,0 +1,10 @@
+package node
+
+type ChanQueue chan struct{}
+
+func MakeChanQueue(n int) ChanQueue {
+	return make(chan struct{}, n)
+}
+
+func (s ChanQueue) acquire() { s <- struct{}{} }
```
yilunzhang: acquire and release make me think of lock rather than channel. Is there a better name for this? or is this naming common?

net/node/node.go
doraemon0: I tested goroutine performance just now. Creating 20480 goroutines and do a simple memory operation(1K memory) in each goroutine will cost 100M memory. Will change it to 20480.

net/node/node.go
```diff
@@ -28,6 +28,7 @@ import (
 
 const (
 	MaxSyncHeaderReq     = 2                // max concurrent sync header request count
```
yilunzhang: Similarly, is 2 too small?

net/node/node.go
```diff
@@ -28,6 +28,7 @@ import (
...
+	MaxMsgChanNum        = 20480            // max goroutine num for message handler
```
yilunzhang: What about 2^N or 10^N haha? You can ignore this 😂

net/node/node.go
doraemon0: As we talked about this offline. const 2 is reasonable.

PR #272, opened 2018-09-07
Name Service

trueinsider:
Signed-off-by: Nikolai Perevozchikov <trueinsider@gmail.com>

To be able to use easy-to-remeber addresses instead of public keys when sending messages I propose name service where anyone can register simple name for their public key.

### Type

### Checklist

### Extra information
**TODO:**
- [x] Test
- [x] Name Restrictions (symbols that can be used, length)
- [x] Only one name per public key
- [x] Name should be unique (exist for only one public key at a time)
- [x] Delete Name
- [x] Owner Verification (only owner should be able to delete)

### DIFF
--- a/core/transaction/payload/RegisterName.go
+++ b/core/transaction/payload/RegisterName.go
@@ -1,0 +1,82 @@
+package payload
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+
+	"github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/common/serialization"
+	. "github.com/nknorg/nkn/errors"
+)
+
+type RegisterName struct {
+	Registrant []byte
+	Name       string
+}
+
+func (a *RegisterName) Data(version byte) []byte {
+	//TODO: implement RegisterName.Data()
+	return []byte{0}
+
+}
+
+func (a *RegisterName) Serialize(w io.Writer, version byte) error {
+	serialization.WriteVarBytes(w, a.Registrant)
+	serialization.WriteVarString(w, a.Name)
+	return nil
+}
+
+func (a *RegisterName) Deserialize(r io.Reader, version byte) error {
+	var err error
+	a.Registrant, err = serialization.ReadVarBytes(r)
+	if err != nil {
+		return NewDetailErr(err, ErrNoCode, "[RegisterName], Registrant Deserialize failed.")
+	}
+	a.Name, err = serialization.ReadVarString(r)
+	if err != nil {
+		return NewDetailErr(err, ErrNoCode, "[RegisterName], Name Deserialize failed.")
+	}
+	return nil
+}
+
+func (a *RegisterName) Equal(b *RegisterName) bool {
+	if !bytes.Equal(a.Registrant, b.Registrant) {
+		return false
+	}
+
+	if a.Name != b.Name {
+		return false
+	}
+
+	return true
+}
+
+func (a *RegisterName) MarshalJson() ([]byte, error) {
+	ra := &RegisterNameInfo{
+		Registrant: common.BytesToHexString(a.Registrant),
+		Name:       a.Name,
+	}
+
+	data, err := json.Marshal(ra)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+func (a *RegisterName) UnmarshalJson(data []byte) error {
+	ra := new(RegisterNameInfo)
+	var err error
+	if err = json.Unmarshal(data, &ra); err != nil {
+		return err
+	}
+
+	a.Registrant, err = common.HexStringToBytes(ra.Registrant)
+	if err != nil {
+		return err
+	}
+	a.Name = ra.Name
+
+	return nil
+}
--- a/core/transaction/transaction.go
+++ b/core/transaction/transaction.go
@@ -32,6 +32,9 @@ const (
 	Prepaid       TransactionType = 0x40
 	Withdraw      TransactionType = 0x41
 	Commit        TransactionType = 0x42
+	RegisterName  TransactionType = 0x50
+	TransferName  TransactionType = 0x51
+	DeleteName    TransactionType = 0x52
 )
 
 type TransactionResult map[Uint256]Fixed64
@@ -205,6 +208,10 @@ func (tx *Transaction) DeserializeUnsignedWithoutType(r io.Reader) error {
 		tx.Payload = new(payload.Withdraw)
 	case Commit:
 		tx.Payload = new(payload.Commit)
+	case RegisterName:
+		tx.Payload = new(payload.RegisterName)
+	case DeleteName:
+		tx.Payload = new(payload.DeleteName)
 	default:
 		return errors.New("[Transaction],invalide transaction type.")
 	}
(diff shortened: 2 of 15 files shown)


trueinsider:
@gdmmx yes, I need to fix that, please see **TODO** above, there is *Only one name per public key*
still work in progress

trueinsider:
@gdmmx I added fix, please look

trueinsider:
@gdmmx thank you, but please don't merge this yet

trueinsider:
I implemented Delete Name and now I have two questions:
1. How should I verify that person can delete only his own name?
2. Is transfer really needed? It shouldn't be possible to transfer name to pub key which already have name for obvious reasons. But even if it doesn't have name yet, someone other can transfer name to someone without even asking him. I imagine there could be cases when this isn't desirable by name recipient.

core/transaction/payload/RegisterName.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/json"
+	"io"
+
+	"github.com/nknorg/nkn/common/serialization"
+	"github.com/nknorg/nkn/crypto"
+	. "github.com/nknorg/nkn/errors"
+)
+
+type RegisterName struct {
+	Registrant *crypto.PubKey
```
doraemon0: Looks use programHash here is more reasonable, since raw public key is heavy for database.

core/transaction/transaction.go
```diff
@@ -339,6 +346,30 @@ func (tx *Transaction) GetProgramHashes() ([]Uint160, error) {
 			return nil, NewDetailErr(err, ErrNoCode, "[Transaction - BookKeeper], GetProgramHashes ToCodeHash failed.")
 		}
 		hashs = append(hashs, astHash)
+	case RegisterName:
+		registrant := tx.Payload.(*payload.RegisterName).Registrant
```
doraemon0: If use programhash in payload, these code could be moved to transaction construction layer.

core/transaction/validator.go
```diff
@@ -319,10 +320,17 @@ func CheckTransactionPayload(txn *Transaction) error {
 			return errors.New("asset is not enough")
 		}
 	case *payload.RegisterName:
-		name := pld.Name
-		registrant, err := Store.GetName(name)
+		name, err := Store.GetName(pld.Registrant)
+		if name != nil {
+			return errors.New(fmt.Sprintf("pubKey %+v already has registered name %s", pld.Registrant, *name))
```
yilunzhang: Shall we restrict to one name per pubkey? Personally I think it's reasonable to allow multiple name to be mapped to the same pubkey (alias), as long as each name is only mapped to one pubkey.

db/prefix.go
```diff
@@ -25,6 +25,10 @@ const (
 	SYS_CurrentBlock      DataEntryPrefix = 0x40
 	SYS_CurrentBookKeeper DataEntryPrefix = 0x42
 
+	// NAME
+	NS_Registrant DataEntryPrefix = 0x50
```
doraemon0: Why need two prefixes? Is it enough if only use an item in database, such as key: Prefix + name, value: programhash?

doraemon0:
@trueinsider 
For your first question:
After RegisterName to blockchain, the owner(public key or programhash)of the name have been recorded in database. If someone want to delete the name, need to provide correct signature DeleteName transaction. For the code, when you add code to GetProgramHashes() for transaction, you have done it.

yilunzhang:
Shall we restrict to one name per pubkey? Personally I think it's reasonable to allow multiple name to be mapped to the same pubkey (alias), as long as each name is only mapped to one pubkey.

core/transaction/payload/RegisterName.go
trueinsider: @doraemon0 We must use public key, because name is resolved to public key and relay is done with using public key (see `server/relay.go`). We can't get public key from program hash.

db/prefix.go
trueinsider: We need to resolve fast not only name -> public key but public key -> name too to check that this public key doesn't have name registered already

trueinsider:
@yilunzhang Why do you need many names for single public key? You can still use identifier with name, same as with public key (part before `.` symbol)

doraemon0:
@trueinsider "So I could just check Owner Verification (only owner should be able to transfer or delete) as done?"  I think, yes.

core/transaction/payload/RegisterName.go
doraemon0: How about encoded public key?

trueinsider: @doraemon0 I think you should always use full public key because it's used as key in Chord always in full form. And Chord doesn't know anything about it being public key, so for Chord it's just a bunch of bytes. @yilunzhang please correct me if I'm wrong

db/prefix.go
trueinsider: Can resolve?

trueinsider:
@yilunzhang @doraemon0 @gdmmx So, as I see it, the only things left is to implement name restrictions and to test that it works as expected?

yilunzhang:
@trueinsider Some people want to be recognized as multiple alias. For example, I use 2 names/alias: yilun for formal case, and skysniper for informal case.

trueinsider:
@yilunzhang We could always lift restrictions later but we can't do other way. If public keys already have several aliases we can't enable only one per public key. I think we should do more restrictive version first and then in future think carefully if restrictions should be lifted. For your case you could just use two separate public keys (won't you want to also separate your formal and informal assets?)

yilunzhang:
@trueinsider That's a very good point. We can always make it less restricted later but not the other way.

core/transaction/payload/RegisterName.go
doraemon0: Chord is using encoded public key. When need to verify something, such as signature chain, decoded it to raw public key first.

core/transaction/transaction.go
doraemon0: Need to do transform from encoded public key to program hash. Maybe need a function like following:
```
func CreateSignatureRedeemScriptWithEncodedPublicKey(encodedPublicKey []byte) ([]byte, error) {
	sb := pg.NewProgramBuilder()
	sb.PushData(encodedPublicKey)
	sb.AddOp(vm.CHECKSIG)
	return sb.ToArray(), nil
}
```

trueinsider:
@doraemon0 I already changed so that DB saves encoded pubkey as []byte, why need to change signature? crypto.Pubkey is easier to use than []byte. Also look at RegisterAsset please, it uses crypto.Pubkey too

core/transaction/transaction.go
trueinsider: Done. Can resolve?

core/transaction/payload/RegisterName.go
trueinsider: Done. Can resolve?

core/transaction/validator.go
```diff
@@ -320,6 +321,14 @@ func CheckTransactionPayload(txn *Transaction) error {
 			return errors.New("asset is not enough")
 		}
 	case *payload.RegisterName:
+		match, err := regexp.MatchString("([a-z]{8,12})", pld.Name)
```
yilunzhang: Why don't we allow A-Z and shorter/longer name?

doraemon0:
@trueinsider looks good to me now, thanks for your hard work!

core/transaction/validator.go
trueinsider: Same reason why we don't allow several names to one public key :) A-Z would mean than names are case sensitive (it's a big decision) and min/max length chosen are just reasonable defaults that can be changed later.

trueinsider:
I tested and everything works. If everything is ok from your side then I think this can be merged.

trueinsider:
Hooray! 🎉

PR #284, opened 2018-09-12
CORS Access Error Fix

98farhan94:
Access control headers added to response to fix the preflight test issue because of CORS leading to network error over TLS query.

screen shot 2018-09-12 at 9 41 20 am

Actual Error: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

- [x ] Bug fix: Link to the issue

### Checklist

### Extra information
RPC calls were failing over TLS because of missing access headers.

### DIFF
--- a/api/httpjson/RPCserver.go
+++ b/api/httpjson/RPCserver.go
@@ -55,9 +55,6 @@ func NewServer(node protocol.Noder, wallet vault.Wallet) *RPCServer {
 }
 
 func (s *RPCServer) write(w http.ResponseWriter, data []byte) {
-	w.Header().Add("Access-Control-Allow-Headers", "Content-Type")
-	w.Header().Set("content-type", "application/json;charset=utf-8")
-	w.Header().Set("Access-Control-Allow-Origin", "*")
 	w.Write(data)
 }
 
--- a/api/httpjson/RPCserver.go
+++ b/api/httpjson/RPCserver.go
@@ -57,11 +57,12 @@ func NewServer(node protocol.Noder, wallet vault.Wallet) *RPCServer {
 //this is the funciton that should be called in order to answer an rpc call
 //should be registered like "http.HandleFunc("/", httpjsonrpc.Handle)"
 func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
+	s.mainMux.RLock()
+	defer s.mainMux.RUnlock()
+	//CORS headers
 	w.Header().Add("Access-Control-Allow-Headers", "Content-Type")
 	w.Header().Set("content-type", "application/json;charset=utf-8")
 	w.Header().Set("Access-Control-Allow-Origin", "*")
-	s.mainMux.RLock()
-	defer s.mainMux.RUnlock()
 	//JSON RPC commands should be POSTs
 	if r.Method != "POST" {
 		if s.mainMux.defaultFunction != nil {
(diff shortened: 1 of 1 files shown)


yilunzhang:
Should we make CORS optional in case some people don't want their node to be access cross site?

98farhan94:
@yilunzhang You are right, it can become a potential attack vector. It will better for security to have it disabled by default.

In my case I am using the `<domain>` instead of `*` (allow all) to make sure only my website can access it.

doraemon0:
The CORS code already existed:
https://github.com/nknorg/nkn/blob/master/api/httpjson/RPCserver.go#L57

It is being called at line:
https://github.com/nknorg/nkn/blob/master/api/httpjson/RPCserver.go#L133
https://github.com/nknorg/nkn/blob/master/api/httpjson/RPCserver.go#L150

Please help to check the root cause.

98farhan94:
@doraemon0 I did notice the presence of CORS in that line, it just wasn't being called. It seems CORS "preflight" request is an `OPTIONS` request instead of `POST` ([Mozilla Developer](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request)) .

Could it be that because it's an OPTIONS request it gets deferred and does not use s.write function, and goes to the default request instead?
https://github.com/nknorg/nkn/blob/master/api/httpjson/RPCserver.go#L57

Am I right to think this is the function that gets called on preflight check? I could be wrong, I haven't worked with RPC before.
https://github.com/nknorg/nkn/blob/master/api/httpjson/RPCserver.go#L162

yilunzhang:
It's not ideal to have both places set CORS, and from the code they should both work or not work. Just to make sure: could you please do 2 tests, one with CORS only in write (without your patch), and one with only CORS only in handle (remove the existing ones in write)? If only one works, we should remove the other.

api/httpjson/RPCserver.go
```diff
@@ -64,6 +64,9 @@ func (s *RPCServer) write(w http.ResponseWriter, data []byte) {
 //this is the funciton that should be called in order to answer an rpc call
 //should be registered like "http.HandleFunc("/", httpjsonrpc.Handle)"
 func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
+	w.Header().Add("Access-Control-Allow-Headers", "Content-Type")
+	w.Header().Set("content-type", "application/json;charset=utf-8")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
doraemon0: I prefer move these lines to the read lock.

98farhan94:
Without the fix:
screen shot 2018-09-13 at 2 11 01 pm

With the fix and write code removed:
screen shot 2018-09-13 at 2 14 45 pm

So I guess we can just remove the original code.

api/httpjson/RPCserver.go
98farhan94: Done.

api/httpjson/RPCserver.go
```diff
@@ -55,9 +55,6 @@ func NewServer(node protocol.Noder, wallet vault.Wallet) *RPCServer {
 }
 
 func (s *RPCServer) write(w http.ResponseWriter, data []byte) {
```
doraemon0: This function is no longer required. Please remove this function and change `s.write(w, data)` to `w.Write(data)` is better.

api/httpjson/RPCserver.go
98farhan94: Updated and also removed unused function.

api/httpjson/RPCserver.go
98farhan94: Did not realise, done!

PR #295, opened 2018-09-16
Fixed RPCServer throwing errors on CORS preflight

CrackDavid:
Signed-off-by: Christian Busch <kontakt@christian-busch.com>

Some Servers sent a CORS preflight to see if the endpoint can actually understand the CORS-protocol. This is a OPTIONS request that causes the RPCServer to run into an error (https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request). 
Made RPCServer.go only to listen for POST-requests and allowed needed header-requests.

- [x] Bug fix

### Checklist

### Extra information
none

### DIFF
--- a/api/httpjson/RPCserver.go
+++ b/api/httpjson/RPCserver.go
@@ -60,91 +60,69 @@ func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
 	s.mainMux.RLock()
 	defer s.mainMux.RUnlock()
 	//CORS headers
-	w.Header().Add("Access-Control-Allow-Headers", "Content-Type")
+	w.Header().Add("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept")
 	w.Header().Set("content-type", "application/json;charset=utf-8")
 	w.Header().Set("Access-Control-Allow-Origin", "*")
-	//JSON RPC commands should be POSTs
-	if r.Method != "POST" {
-		if s.mainMux.defaultFunction != nil {
-			log.Info("HTTP JSON RPC Handle - Method!=\"POST\"")
-			s.mainMux.defaultFunction(w, r)
-			return
-		} else {
-			log.Warn("HTTP JSON RPC Handle - Method!=\"POST\"")
+	if r.Method == "POST" {
+		//read the body of the request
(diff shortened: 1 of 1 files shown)


api/httpjson/RPCserver.go
```diff
@@ -60,91 +60,69 @@ func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
...
+			w.Write(data)
+		} else {
+			//if the function does not exist
+			log.Warn("HTTP JSON RPC Handle - No function to call for ", request["method"])
+			data, err := json.Marshal(map[string]interface{}{
 				"jsonrpc": "2.0",
 				"error": map[string]interface{}{
-					"code":    -errcode,
-					"message": common.ErrMessage[errcode],
+					"code":    -32601,
```
dreamfly281: It is better using some meaning errcode instead of the number.

CrackDavid: These lines weren't changed from the original file (https://github.com/nknorg/nkn/blob/6a62178497c8e922b4cba2660103064343ade501/api/httpjson/RPCserver.go#L137) - although git tells you so :)
I changed the code according to the responses above so it is more dynamic. Can you please have a look into that?

api/httpjson/RPCserver.go
```diff
@@ -60,91 +60,70 @@ func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
...
+			}
+			if err != nil {
+				log.Error("HTTP JSON RPC Handle - json.Marshal: ", err)
+				return
+			}
+			w.Write(data)
+		} else {
+			//if the function does not exist
+			log.Warn("HTTP JSON RPC Handle - No function to call for ", request["method"])
+			response := function(s, request["params"].(map[string]interface{}))
```
yilunzhang: This line can be removed

api/httpjson/RPCserver.go
```diff
@@ -60,91 +60,70 @@ func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
...
+			errcode := response["error"].(common.ErrCode)
```
yilunzhang: This line can be changed to
errcode := common.INVALID_METHOD

PR #331, opened 2018-10-09
validate tx with other txs inside pool/block

trueinsider:
Transactions should be validated with other transactions in pool/block when transaction is received or block is proposed/received/voted. Unify validation logic.

### Type

### Checklist

### DIFF
--- a/core/transaction/pool/pool.go
+++ b/core/transaction/pool/pool.go
@@ -8,7 +8,6 @@ import (
 	"github.com/nknorg/nkn/common"
 	"github.com/nknorg/nkn/core/ledger"
 	. "github.com/nknorg/nkn/core/transaction"
-	"github.com/nknorg/nkn/core/transaction/payload"
 	. "github.com/nknorg/nkn/errors"
 	"github.com/nknorg/nkn/por"
 	"github.com/nknorg/nkn/util/log"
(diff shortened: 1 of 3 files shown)


core/transaction/pool/pool.go
```diff
@@ -43,14 +38,18 @@ func (tp *TxnPool) AppendTxnPool(txn *Transaction) ErrCode {
 		log.Info("Transaction verification failed", txn.Hash())
 		return errCode
 	}
+	var txnList []*Transaction
+	for _, tx := range tp.txnList {
+		txnList = append(txnList, tx)
+	}
+	if errCode := VerifyTransactionWithBlock(txnList); errCode != ErrNoError {
```
NoelBright: this will cost a lot of time to verify a transaction, when the txpool is very large.

trueinsider: do you have ideas how this can be fixed? I thought to make `VerifyTransactionWithBlock` accept map instead of array but it's called also in `core/ledger/validator.go` and there is array

trueinsider: ok, I think I fixed this @NoelBright

NoelBright:
VerifyTransactionWithBlock is used to verify transactions in a block. Using it to verify txPool is unsuitable. For example  the verification of IssueAsset in VerifyTransactionWithBlock is not necessary.
I suggest that  using a new function to replace VerifyTransactionWithBlock.

trueinsider:
@NoelBright previous problem is fixed already.
`VerifyTransactionWithBlock` is not my function, it already existed. And it wasn't used anywhere in the code. But logic that was used in `AppendTxnPool` in it's place is exactly the same from what I see, including verification of `IssueAsset`. So I didn't really changed behavior of `AppendTxnPool`, I just got rid of duplicate code.
Also I added call to `VerifyTransactionWithBlock` when new block is proposed/received and `IssueAsset` verification is needed there too, because that block could've been received from malicious node.

trueinsider:
@NoelBright @yilunzhang I optimized code so now it's linear time for all cases.

PR #334, opened 2018-10-12
verify duplicate name service operations in a block/pool

trueinsider:
Right now name/registrant verification in name service happens only with ledger but not with the pool, because of this there could be two conflicting txs in a block that would cause problems when trying to persist this block.

### Type

### Checklist

### DIFF
--- a/core/transaction/validator.go
+++ b/core/transaction/validator.go
@@ -75,6 +75,8 @@ func VerifyTransactionWithBlock(iterator Iterator) ErrCode {
 	txnlist := make(map[Uint256]struct{}, 0)
 	txPoolInputs := make(map[string]struct{}, 0)
 	issueSummary := make(map[Uint256]Fixed64, 0)
+	registeredNames := make(map[string]struct{}, 0)
+	nameRegistrants := make(map[string]struct{}, 0)
 	//start check
 	return iterator.Iterate(func(txn *Transaction) ErrCode {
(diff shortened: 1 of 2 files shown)


core/transaction/validator.go
```diff
@@ -129,6 +131,31 @@ func VerifyTransactionWithBlock(iterator Iterator) ErrCode {
...
+			nameRegistrants[registrant] = struct{}{}
+		case DeleteName:
+			namePayload := txn.Payload.(*payload.RegisterName)
+
+			registrant := BytesToHexString(namePayload.Registrant)
+			if _, ok := nameRegistrants[registrant]; ok {
+				log.Warn("[VerifyTransactionWithBlock], duplicate registrant exist in block.")
+				return ErrDuplicateName
+			}
+			nameRegistrants[registrant] = struct{}{}
```
yilunzhang: Does this mean we won't allow someone to delete and register a name in the same block?

trueinsider: yes, because there is no guarantee that they will be persisted in the correct order

yilunzhang: OK sounds reasonable

yilunzhang:
Looks good to me. @dreamfly281 @NoelBright Probably you can take a quick look and then we can merge this

trueinsider:
@dreamfly281 done

yilunzhang:
@dreamfly281

PR #381, opened 2018-11-21
Limit transaction inputs/outputs NOT over MaxRefIndex

gdmmx:
- Limit transaction inputs/outputs NOT over max of ReferTxOutputIndex
- Complemented ErrCode String
- Refine code of IsDoubleSpend routine

### Checklist

### Extra information


core/transaction/input.go
```diff
@@ -19,6 +19,9 @@ type TxnInput struct {
 	ReferTxOutputIndex uint16
 }
 
+// MaxRefIndex is the max of TxnInput.ReferTxOutputIndex
+const MaxRefIndex = ^uint16(0)
```
yilunzhang: Should we use math.MaxUint16 instead?

core/transaction/validator.go
```diff
@@ -178,6 +181,12 @@ func VerifyTransactionWithLedger(Tx *Transaction) ErrCode {
 	return ErrNoError
 }
 
+// CheckInputOutputLen validate len(inputs/outputs) less than MAX of ReferTxOutputIndex
+func CheckInputOutputTooLong(tx *Transaction) bool {
+	// FIXME: assume count of Inputs in a Transaction never over 2147483648
```
yilunzhang: I think we are good since len() returns a int type anyway

db/store.go
```diff
@@ -255,30 +255,28 @@ func (cs *ChainStore) IsTxHashDuplicate(txhash Uint256) bool {
...
+			log.Error(err)
 			return true
 		}
+		for _, v := range unspents {
+			if v == input.ReferTxOutputIndex {
+				continue NEXT // found in unspents
+			}
+		}
+		log.Warningf("Transaction %s refer a spent Output [%x:%d]", tx.Hash(), txhash, input.ReferTxOutputIndex)
+		return true // ReferTxOutputIndex not in unspents
```
yilunzhang: Will it return true if tx input is referring to some non existing unspents (like unspents is empty)? This is still invalid tx but just sounds like another invalid type

core/transaction/input.go
gdmmx: I suggest don't instead of MaxUint16. Because we can't assumed type of ReferTxOutputIndex always is uint16.

core/transaction/validator.go
gdmmx: 2147483648 will be regarded as negative on 32 bits platform. And int of golang will be either 32 or 64 bits depend on implementation-specific.
I am not handle this corner case since transaction contain so much Inputs/Outputs would be a huge msg can't broadcast on network at all.

db/store.go
gdmmx: There are another verification CheckTransactionBalance() to check it already. So I don't repetitive work in this routine.

core/transaction/validator.go
gdmmx: I mean that len(tx.Inputs) will return a negative cause bool ( len(tx.Inputs) > 65535 ) as false.

core/transaction/input.go
gdmmx: My expected solution is 
```
type IdxType uint16
ReferTxOutputIndex IdxType
MaxRefIndex = 1<<(sizeof(IdxType)*8) -1
```
But golang can't unsafe.Sizeof a Type without instantiated...
Have you any suggestions?

yilunzhang: I think it's enough to define MaxRefIndex as int, e.g.
const MaxRefIndex = 12345
because it's always compared with int (return of len)

gdmmx: Change the type of ReferTxOutputIndex is backward incompatible update...

gdmmx: @yilunzhang I submitted a new change 
182ce51, type define MaxRefIndex as int with initialized via math.MaxUint16.

yilunzhang:
Looks good to me

PR #384, opened 2018-11-27
Will try nat by default and change nat flag to no-nat

yilunzhang:

### Checklist

### Extra information


util/config/config.go
```diff
@@ -95,38 +96,38 @@ func Init() error {
 		Parameters.IncrementPort()
 	}
 
-	if Parameters.NAT {
+	if Parameters.NAT && !SkipNAT {
```
gdmmx: @yilunzhang The config.Parameters.NAT default is false. This condition will never true.

util/config/config.go
```diff
@@ -95,38 +96,38 @@ func Init() error {
...
-		if err != nil {
-			return err
-		}
-		log.Printf("Mapped external port %d to internal port %d", externalPort, internalPort)
-
-		externalPort, internalPort, err = nat.AddPortMapping("tcp", int(Parameters.HttpJsonPort), int(Parameters.HttpJsonPort), "nkn", 10*time.Second)
-		if err != nil {
-			return err
+		if err == nil {
+			log.Printf("Found %s gateway", nat.Type())
```
dreamfly281: How about use NKN own log mechanism like  log.Info or log.Debug to keep the log output consistent?

yilunzhang: Because our log imports config to get log settings so using it would cause a circular dependency. In the future we should change the code structure and use something like setlogger to config logger like nnet does it now.

PR #394, opened 2018-12-09
initial pubsub implementation

trueinsider:
Ability to subscribe to topics and publish messages to them.

### Type
- [X] Feature (Non-breaking change)

### Checklist
- [X] Read the [CONTRIBUTION guidelines](https://github.com/nknorg/nkn#contributing).
- [X] All the tests are passing after the introduction of new changes.
- [X] Code has been written according to [NKN-Golang-Style-Guide](https://github.com/nknorg/nkn/wiki/NKN-Golang-Style-Guide)

### Extra information
Wallet SDK PR: https://github.com/nknorg/nkn-wallet-js/pull/15
Client SDK PR: https://github.com/nknorg/nkn-client-js/pull/34

### DIFF
--- a/core/transaction/payload/Subscribe.go
+++ b/core/transaction/payload/Subscribe.go
@@ -1,0 +1,111 @@
+package payload
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+
+	"github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/common/serialization"
+	. "github.com/nknorg/nkn/errors"
+	"github.com/nknorg/nkn/util/address"
+)
+
+type Subscribe struct {
+	Subscriber []byte
+	Identifier string
+	Topic      string
+	Duration   uint32
+}
+
+func (s *Subscribe) Data(version byte) []byte {
+	//TODO: implement Subscribe.Data()
+	return []byte{0}
+
+}
+
+func (s *Subscribe) Serialize(w io.Writer, version byte) error {
+	serialization.WriteVarBytes(w, s.Subscriber)
+	serialization.WriteVarString(w, s.Identifier)
+	serialization.WriteVarString(w, s.Topic)
+	serialization.WriteUint32(w, s.Duration)
+	return nil
+}
+
+func (s *Subscribe) Deserialize(r io.Reader, version byte) error {
+	var err error
+	s.Subscriber, err = serialization.ReadVarBytes(r)
+	if err != nil {
+		return NewDetailErr(err, ErrNoCode, "[Subscribe], Subscriber Deserialize failed.")
+	}
+	s.Identifier, err = serialization.ReadVarString(r)
+	if err != nil {
+		return NewDetailErr(err, ErrNoCode, "[Subscribe], Identifier Deserialize failed.")
+	}
+	s.Topic, err = serialization.ReadVarString(r)
+	if err != nil {
+		return NewDetailErr(err, ErrNoCode, "[Subscribe], Topic Deserialize failed.")
+	}
+	s.Duration, err = serialization.ReadUint32(r)
+	if err != nil {
+		return NewDetailErr(err, ErrNoCode, "[Subscribe], Duration Deserialize failed.")
+	}
+	return nil
+}
+
+func (s *Subscribe) Equal(s2 *Subscribe) bool {
+	if !bytes.Equal(s.Subscriber, s2.Subscriber) {
+		return false
+	}
+
+	if s.Identifier != s2.Identifier {
+		return false
+	}
+
+	if s.Topic != s2.Topic {
+		return false
+	}
+
+	if s.Duration != s2.Duration {
+		return false
+	}
+
+	return true
+}
+
+func (s *Subscribe) SubscriberString() string {
+	return address.MakeAddressString(s.Subscriber, s.Identifier)
+}
+
+func (s *Subscribe) MarshalJson() ([]byte, error) {
+	si := &SubscribeInfo{
+		Subscriber: common.BytesToHexString(s.Subscriber),
+		Identifier: s.Identifier,
+		Topic:      s.Topic,
+		Duration:   s.Duration,
+	}
+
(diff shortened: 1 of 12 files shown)


NoelBright:
Lost roll back function when storing Subscription & Unsubsciption in DB.

NoelBright:
I will test these codes in testing network.

trueinsider:
@dreamfly281 @NoelBright please look again, I've made some changes
There is now duration for subscription, subscriptions limit per topic and rollback

dreamfly281:
> @dreamfly281 @NoelBright please look again, I've made some changes
> There is now duration for subscription, subscriptions limit per topic and rollback

The code looks good from my side, NoelBright will merge it into testbed to run some test cases before merged into master branch, thanks

db/store.go
```diff
@@ -738,6 +852,14 @@ func (cs *ChainStore) persist(b *Block) error {
 	// BATCH PUT VALUE
 	cs.st.BatchPut(bhash.Bytes(), hashWriter.Bytes())
 
+	expiredKeys := cs.GetExpiredKeys(b.Header.Height)
```
NoelBright: The element of expiredKeys does not contain prefix ”SYS_ExpireKey + height". But BatchDelete need the full key that contain this prefix to delete expired keys.

trueinsider: Actually, I should delete two keys, both with prefix and without. Since expired keys contains full key names that should be deleted at specified block.

NoelBright: cannot use subscribePayload.Duration (type uint32) as type uint16 in argument to cs.Subscribe

core/transaction/payload/Subscribe.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"io"
+	"strings"
+
+	"github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/common/serialization"
+	. "github.com/nknorg/nkn/errors"
+)
+
+type Subscribe struct {
```
yilunzhang: FYI, all data structure will be migrated to protobuf in the future, do you want to migrate later (like most txn now) or straightly start with protobuf (like sigchain)?

trueinsider: I'll migrate later for consistency sake

NoelBright:
[test.log](https://github.com/nknorg/nkn/files/2689764/test.log)

when I add multiple subscribe txs in one block, only one of these subscribers can be removed from db.
look the test.log for detail information.

the subscribers below are not removed from db .I can get them from RPC.
{
    "id": 1,
    "jsonrpc": "2.0",
    "result": [
        "002.03492ecaf7454fcd81c481fead3a146950a2ff6fa8247a8b845883096b4987a9fe",
        "003.03492ecaf7454fcd81c481fead3a146950a2ff6fa8247a8b845883096b4987a9fe",
        "004.03492ecaf7454fcd81c481fead3a146950a2ff6fa8247a8b845883096b4987a9fe"
    ]
}

core/transaction/validator.go
```diff
@@ -348,6 +379,33 @@ func CheckTransactionPayload(txn *Transaction) error {
...
 			return errors.New(fmt.Sprintf("no name registered for pubKey %+v", pld.Registrant))
 		}
+	case *payload.Subscribe:
+		duration := pld.Duration
+		if duration > MaxSubscriptionDuration {
+			return errors.New(fmt.Sprintf("subscription duration %d can't be bigger than %d", duration, MaxSubscriptionDuration))
+		}
+
+		topic := pld.Topic
+		match, err := regexp.MatchString("([a-z]{8,12})", topic)
```
yilunzhang: should we give a little bit more flexibility in topic name? like a hyphen and a wider length limit?

trueinsider: no problem, just say the values

yilunzhang:
I have one concern here: the client address scheme (identifier.pubkey) is heavily coupled into the code now, so if we change the scheme, lots of functions needs to be updated. Currently the only place that relies on the scheme is ParseClientAddress (in util/address/address.go), and identifier is never explicitly passed because it's not a basic variable. Whole client address and public key are basic variable because they are needed for any scheme we use. I think it's better to use the current approach so that the pub/sub code is decoupled from the client address scheme. What do you think?

core/transaction/validator.go
yilunzhang: This is Google Cloud Pub/Sub's restrictions for reference. It's url friendly and thus quite safe.

- Start with a letter
- Contain between 3 and 255 characters
- Contain only the following characters:
  - Letters: [A-Za-z]
  - Numbers: [0-9]
  - Dashes: -
  - Underscores: _
  - Periods: .
  - Tildes: ~
  - Plus signs: +
  - Percent signs: %

core/transaction/payload/Subscribe.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (s *Subscribe) UnmarshalJson(data []byte) error {
+	si := new(SubscribeInfo)
+	var err error
+	if err = json.Unmarshal(data, &si); err != nil {
+		return err
+	}
+
+	substrings := strings.Split(si.Subscriber, ".")
```
yilunzhang: What about move the client address encode/decode part into somewhere (e.g. util/address/address.go) so that all address scheme code are in the same module and other (future) scheme-related code can also use it?

trueinsider: done

NoelBright:
looks good to me.

NoelBright:
I think the "move address encoding to address.go"  could be another commit.

trueinsider:
@NoelBright it's directly related to pubsub and doesn't change anything except pubsub. There'll be one more commit to change topic name restrictions.

core/transaction/validator.go
trueinsider: done

yilunzhang:
Looks good to me.

NoelBright:
It passes my test cases.

PR #419, opened 2019-01-17
count node relay data

TDCQZD:

### Checklist

### Extra information

### DIFF
--- a/api/common/interfaces.go
+++ b/api/common/interfaces.go
@@ -5,7 +5,6 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"math"
-	"sync/atomic"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/nknorg/nkn/common"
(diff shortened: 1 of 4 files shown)


net/node/node.go
```diff
@@ -12,6 +12,8 @@ import (
 	nnetpb "github.com/nknorg/nnet/protobuf"
 )
 
+var NodeRelayMessage uint64 = 0
```
yilunzhang: It's better make it a (private) field of LocalNode structure because conceptually it makes more sense (and works if we have multiple local nodes for example)

api/common/interfaces.go
```diff
@@ -129,6 +130,14 @@ func getLatestBlockHeight(s Serverer, params map[string]interface{}) map[string]
 	return respPacking(ledger.DefaultLedger.Blockchain.BlockHeight, SUCCESS)
 }
 
+func getNodeRelayMessageCount(s Serverer, params map[string]interface{}) map[string]interface{} {
```
yilunzhang: There is no need to add a new API endpoint for it. It would be good if people can get this result using `./nknc info -s`. To do that you just need to put it in LocalNode MarshalJSON method, together with height and version.

net/node/message.go
```diff
@@ -202,6 +203,7 @@ func (localNode *LocalNode) remoteMessageRouted(remoteMessage *nnetnode.RemoteMe
 				log.Errorf("Marshal new relay msg body error: %v", err)
 				return nil, nil, nil, false
 			}
+			atomic.AddUint64(&NodeRelayMessage, 1)
```
yilunzhang: As suggested by the Go doc: https://golang.org/pkg/sync/atomic/ it's better not to use sync/atomic directly. If you make it a LocalNode field, you can use the existing sync.RWMutex in LocalNode to handle atomic read/write.

yilunzhang:
@TDCQZD 
One tiny thing: it's better to run gofmt before commit. Some IDE can do this for you

PR #426, opened 2019-01-20
add meta information to subscribers

trueinsider:
Signed-off-by: Nikolai Perevozchikov <insider@nkn.org>

Ability to add custom meta information to subscriber when subscribing

### Type

### Checklist

### DIFF
--- a/core/transaction/builder.go
+++ b/core/transaction/builder.go
@@ -184,18 +184,19 @@ func NewDeleteNameTransaction(registrant []byte, name string) (*Transaction, err
 	}, nil
 }
 
-func NewSubscribeTransaction(subscriber []byte, identifier string, topic string, bucket uint32, duration uint32) (*Transaction, error) {
+func NewSubscribeTransaction(subscriber []byte, identifier string, topic string, bucket uint32, duration uint32, meta string) (*Transaction, error) {
 	SubscribePayload := &payload.Subscribe{
 		Subscriber: subscriber,
 		Identifier: identifier,
 		Topic: topic,
 		Bucket: bucket,
 		Duration: duration,
+		Meta: meta,
 	}
 
 	return &Transaction{
 		TxType:  Subscribe,
-		PayloadVersion: 1,
+		PayloadVersion: 2,
 		Payload: SubscribePayload,
 		Attributes: []*TxnAttribute{
 			{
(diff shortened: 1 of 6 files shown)


core/transaction/payload/Subscribe.go
```diff
@@ -33,6 +34,9 @@ func (s *Subscribe) Serialize(w io.Writer, version byte) error {
 		serialization.WriteUint32(w, s.Bucket)
 	}
 	serialization.WriteUint32(w, s.Duration)
+	if version == 2 {
```
yilunzhang: I think we should change version == to version >= especially for all the version == 1

yilunzhang:
Looks good to me now

core/transaction/builder.go
```diff
@@ -184,18 +184,19 @@ func NewDeleteNameTransaction(registrant []byte, name string) (*Transaction, err
...
 		Bucket: bucket,
 		Duration: duration,
+		Meta: meta,
 	}
 
 	return &Transaction{
 		TxType:  Subscribe,
-		PayloadVersion: 1,
+		PayloadVersion: 2,
 		Payload: SubscribePayload,
```
dreamfly281: any other place to define or use the version?

core/transaction/payload/Subscribe.go
```diff
@@ -50,7 +54,7 @@ func (s *Subscribe) Deserialize(r io.Reader, version byte) error {
 	if err != nil {
 		return NewDetailErr(err, ErrNoCode, "[Subscribe], Topic Deserialize failed.")
 	}
-	if version == 1 {
+	if version >= 1 {
```
dreamfly281: I'd  prefer describe the difference of these two version, if it incompatible, we'd consider drop the lower version consider we still in testnet instead of maintain "old" version

core/transaction/builder.go
trueinsider: version is also specified in core/transaction/payload/Subscribe.go and SDK if I understand your question correctly

core/transaction/payload/Subscribe.go
trueinsider: It would be incompatible and would require ledger reset, so probably it's better to leave version for now and drop lower versions together with balance/25519

yilunzhang: The version number should just be a workaround now. Starting from balance model, block and transactions are serialized and deserialized through protobuf and should maintain this compatibility automatically.

PR #442, opened 2019-02-17
Change AddPortMapping timeout overflow to int

michaelotis:
UPnP would execute to my router and then subsequently be removed, replacing the calculation of 10*time.second with just 10, the UPnP mapping completes and is not removed. Allowing users to not have have to forward ports if they have UPnP working.

### Checklist

### Extra information

### DIFF
--- a/util/config/config.go
+++ b/util/config/config.go
@@ -164,19 +164,19 @@ func (config *Configuration) addPortMapping(nat gonat.NAT) error {
 		return err
 	}
 
-	externalPort, internalPort, err := nat.AddPortMapping(transport.GetNetwork(), int(config.NodePort), int(config.NodePort), "nkn", 10*time.Second)
+	externalPort, internalPort, err := nat.AddPortMapping(transport.GetNetwork(), int(config.NodePort), int(config.NodePort), "nkn", 10)
 	if err != nil {
 		return err
 	}
 	log.Printf("Mapped external port %d to internal port %d", externalPort, internalPort)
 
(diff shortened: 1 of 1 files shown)


yilunzhang:
The changes is essentially using 0 as timeout since timeout is divided by time.Second (1000000000) before using: https://github.com/nknorg/go-nat/blob/master/upnp.go#L170

I tried it on Apple Airport and it didn't work

yilunzhang:
Does it work for you if you change 10*time.Second to 365*24*time.Hour?

PS: Apple Airport uses NAT-PMP instead of UPnP

michaelotis:
> Does it work for you if you change 10 * time.Second to 365 * 24 * time.Hour?
> 
> PS: Apple Airport uses NAT-PMP instead of UPnP

No, it didn't work. Seems odd since in the standard go-nat example dir, it uses 60 as an example timeout time.

	eport, iport, err := nat.AddPortMapping("tcp", 3081, 3080, "http", 60)

yilunzhang:
In golang type is after variable name, so the type is time.Duration, which is int64, so overflow is not a problem

michaelotis:
Also to follow up about router type, it's a Trendnet running dd-wrt

yilunzhang:
Let us try to find a couple more routers to test

yilunzhang:
@michaelotis #452 makes NATPortMappingTimeout configurable. Once merged, you just need to add `"NATPortMappingTimeout": 0,` in your config.json for your case, while other people can just use the default one.

PR #463, opened 2019-03-14
Add getid4extip RPC

gdmmx:
Replace ipify for more decentralized. 
Split updating as two phases for compatible with online nodes which still old version. 
This is patch of phase 1.

Split updating as two phases for compatible with online nodes which still old version.
1. Add getid4extip RPC to current online nodes.
2. As new version, nknd request extIP via RPC instead of ipify.

### Checklist

### Extra information

### DIFF
--- a/api/common/interfaces.go
+++ b/api/common/interfaces.go
@@ -5,6 +5,8 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"math"
+	"net"
+	"strings"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/nknorg/nkn/common"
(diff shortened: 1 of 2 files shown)


api/common/interfaces.go
```diff
@@ -1154,6 +1157,34 @@ func getTopicBucketsCount(s Serverer, params map[string]interface{}) map[string]
...
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		if strings.LastIndexByte(addr, ':') < 0 { // Just only host, without port
+			host = addr
+		} else {
+			log.Errorf("getID4ExtIP met invalid params %v: %v", addr, err)
+			return respPacking(nil, INVALID_PARAMS)
+		}
+	}
+	id := address.GenChordID(fmt.Sprintf("%s:%d", host, config.Parameters.NodePort))
```
yilunzhang: I think we should use parsed port if addr contains one, if not we should use the default one like here. Otherwise it's not quite convenient in some case (local cluster for example)

gdmmx:
@yilunzhang As offline discussion, I changed RPC from getid4extip to getmyextip.

api/httpjson/RPCserver.go
```diff
@@ -109,6 +109,19 @@ func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
 			w.Write(data)
 			return
 		}
+		// if params["RemoteAddr"] set but empty, used request.RemoteAddr
+		if addr, ok := params["RemoteAddr"]; ok {
```
yilunzhang: Why not set RemoteAddr it if params["RemoteAddr"] is not set?

gdmmx: params["RemoteAddr"] just only significant for getid4extip RPC. It is condition for distinguish getid4extip or others RPC during Handle() routine.

yilunzhang: OK

PR #465, opened 2019-03-22
Add Mining debug info

gdmmx:

Show mining debug info in getnodestate RPC if config.MiningDebug enable, helpful for unlucky miners guys.

### Checklist

### Extra information

### DIFF
--- a/node/localnode.go
+++ b/node/localnode.go
@@ -47,6 +47,7 @@ type LocalNode struct {
 	syncOnce          *sync.Once
 	relayMessageCount uint64    // count how many messages node has relayed since start
 	startTime         time.Time // Time of localNode init
+	proposalSubmitted uint32    // Count of localNode submitted proposal
 }
 
 func (localNode *LocalNode) MarshalJSON() ([]byte, error) {
@@ -66,6 +67,10 @@ func (localNode *LocalNode) MarshalJSON() ([]byte, error) {
 	out["uptime"] = time.Since(localNode.startTime).Truncate(time.Second).Seconds()
 	out["version"] = config.Version
 	out["relayMessageCount"] = localNode.GetRelayMessageCount()
+	if config.Parameters.MiningDebug {
+		out["proposalSubmitted"] = localNode.proposalSubmitted
+		out["currTimeStamp"] = time.Now().Unix()
+	}
 
 	return json.Marshal(out)
 }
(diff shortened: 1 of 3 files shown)


node/localnode.go
```diff
@@ -66,6 +67,10 @@ func (localNode *LocalNode) MarshalJSON() ([]byte, error) {
 	out["uptime"] = time.Since(localNode.startTime).Truncate(time.Second).Seconds()
 	out["version"] = config.Version
 	out["relayMessageCount"] = localNode.GetRelayMessageCount()
+	if config.Parameters.MiningDebug {
+		out["proposalSubmitted"] = localNode.proposalSubmitted
```
yilunzhang: we should use GetProposalSubmitted() instead to prevent race condition

node/localnode.go
```diff
@@ -207,6 +212,14 @@ func (localNode *LocalNode) maybeAddRemoteNode(remoteNode *nnetnode.RemoteNode)
 	return nil
 }
 
+func (localNode *LocalNode) GetProposalSubmitted() uint32 {
+	return localNode.proposalSubmitted
```
yilunzhang: GetProposalSubmitted() and IncrementProposalSubmitted() can be called together, so a lock is needed.

node/localnode.go
```diff
@@ -207,6 +212,14 @@ func (localNode *LocalNode) maybeAddRemoteNode(remoteNode *nnetnode.RemoteNode)
...
+}
+
+func (localNode *LocalNode) IncrementProposalSubmitted() {
+	localNode.proposalSubmitted++
```
yilunzhang: Same as above

node/localnode.go
gdmmx: I consider solidly consistency is unnecessary at here.
localNode.proposalSubmitted just only has one writer in a single goroutine consensus.startProposing(), and never should be in concurrent writing. (concurrent proposal is nonlicet and will be rejected by neighbors)
On the other hand, proposalSubmitted just a reference for evaluate node's propose success rate. Deviation 1 with probability is acceptable. Whatever, getnodeinfo RPC should not influenced the performance of consensus routine for avoid potential DDOS attack.

gdmmx: Anyway, I push a commit for make sure atomic R/W localNode.proposalSubmitted without lock.

PR #490, opened 2019-04-25
Add sigChain parser utility and fix logLevel bug

gdmmx:
Add v0.9 utility for parse sigChain from hex byte stream

### Checklist

### Extra information

### DIFF
--- a/util/bytes2sigchain.go
+++ b/util/bytes2sigchain.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"bufio"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/nknorg/nkn/pb"
+)
+
+// Base64ToHex convert base64 string input to hex string output
+func Base64ToHex(in []byte) (out []byte, err error) { // input format should be ':"balabala..."'
+	const prefix, suffix = ":\"", "\""
+	start, end := len(prefix), len(in)-len(suffix) // b64Str len not contain :" prefix and " suffix
+
+	// Decode Base64 string to bin
+	bin := make([]byte, base64.StdEncoding.DecodedLen(end-start))
+	n, err := base64.StdEncoding.Decode(bin, in[start:end])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "base64 decode %s met erro %v\n", in[start:end], err)
+		return []byte{}, err
+	}
+
+	// Encode bin to hex string
+	hexStr := make([]byte, 2*len(bin[:n]))
+	hex.Encode(hexStr, bin[:n])
+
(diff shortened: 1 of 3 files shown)


gdmmx:
@yilunzhang I pushed a new implement base on regexp find&replace base64 string to hex string.
This solution is carefree about pb .proto changed in future.

util/bytes2sigchain.go
```diff
@@ -0,0 +1,90 @@
...
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/nknorg/nkn/pb"
+)
+
+// Base64ToHex convert base64 string input to hex string output
+func Base64ToHex(in []byte) (out []byte, err error) { // input format should be ':"balabala..."'
+	start, end := len(":\""), len(in)
```
yilunzhang: what about
start, end := len(":\""), len(in)-len("\"")
so that we can use end directly? Currently only end-1 is used.

util/bytes2sigchain.go
```diff
@@ -0,0 +1,90 @@
...
+					js := map[string]interface{}{"SigChain": sc, "Submitter": commitTxn.Submitter}
+					if jStr, err = json.Marshal(js); err == nil { // to json string
+						Base64Handler(re, os.Stdout, jStr) // replace all in jStr to stdout
+						os.Stdout.WriteString("\n")
+						continue
+					}
+				}
+			}
+		}
+		fmt.Fprintf(os.Stderr, "%v\n", err) // any error from above
```
yilunzhang: We should probably only exec this line when err != nil, otherwise there are two \n at the end of the result

util/bytes2sigchain.go
gdmmx: I pushed new commit again for better readability of Base64ToHex.

util/bytes2sigchain.go
yilunzhang: Oh you are right.

util/bytes2sigchain.go
```diff
@@ -0,0 +1,91 @@
...
+// Base64ToHex convert base64 string input to hex string output
+func Base64ToHex(in []byte) (out []byte, err error) { // input format should be ':"balabala..."'
+	const prefix, suffix = ":\"", "\""
+	start, end := len(prefix), len(in)-len(suffix) // b64Str len not contain :" prefix and " suffix
+
+	// Decode Base64 string to bin
+	bin := make([]byte, base64.StdEncoding.DecodedLen(end-start))
+	n, err := base64.StdEncoding.Decode(bin, in[start:end])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "base64 decode %s met erro %v\n", in[start:end-1], err)
```
yilunzhang: should this be in[start:end]?

gdmmx: My fault :joy: 
I had updated my commit.

PR #499, opened 2019-05-05
Nano Payment Channels

trueinsider:
Implement simple version of payment channels that works only in a single direction between 2 users.

### Checklist

### Extra information

### DIFF
--- a/chain/db/nanopay.go
+++ b/chain/db/nanopay.go
@@ -1,0 +1,42 @@
+package db
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/common/serialization"
+)
+
+type nanoPay struct {
+	balance   common.Fixed64
+	expiresAt uint32
+}
+
+func (np *nanoPay) Serialize(w io.Writer) error {
+	err := np.balance.Serialize(w)
+	if err != nil {
+		return fmt.Errorf("nanoPay Serialize error: %v", err)
+	}
+
+	err = serialization.WriteUint32(w, np.expiresAt)
+	if err != nil {
+		return fmt.Errorf("nanoPay Serialize error: %v", err)
+	}
+
+	return nil
+}
+
+func (np *nanoPay) Deserialize(r io.Reader) error {
+	err := np.balance.Deserialize(r)
+	if err != nil {
+		return fmt.Errorf("Deserialize nanoPay error: %v", err)
+	}
+
+	np.expiresAt, err = serialization.ReadUint32(r)
+	if err != nil {
+		return fmt.Errorf("Deserialize nanoPay error: %v", err)
+	}
+
+	return nil
+}
--- a/chain/db/stateGenerator.go
+++ b/chain/db/stateGenerator.go
@@ -73,6 +73,26 @@ func (cs *ChainStore) spendTransaction(states *StateDB, txn *transaction.Transac
 		}
 		states.UpdateBalance(donationAddress, Fixed64(genID.RegistrationFee), Addition)
 
+	case pb.NanoPayType:
+		nanoPay := pl.(*pb.NanoPay)
+		pg, err := txn.GetProgramHashes()
+		if err != nil {
+			return err
+		}
+
+		addrRecipient := BytesToUint160(nanoPay.Recipient)
+		nanoPayBalance, _, err := states.GetNanoPay(pg[0], addrRecipient, nanoPay.Nonce)
+		if err != nil {
+			return err
+		}
+		claimAmount := Fixed64(nanoPay.Amount) - nanoPayBalance
+		if err := states.UpdateBalance(pg[0], claimAmount, Subtraction); err != nil {
+			return err
+		}
+		if err := states.UpdateBalance(addrRecipient, claimAmount, Addition); err != nil {
+			return err
+		}
+		states.SetNanoPay(pg[0], addrRecipient, nanoPay.Nonce, Fixed64(nanoPay.Amount), nanoPay.Height+nanoPay.Duration)
 	}
 
 	return nil
@@ -136,16 +156,18 @@ func (cs *ChainStore) generateStateRoot(b *block.Block, needBeCommitted bool) (*
 		for _, txn := range b.Transactions {
 			cs.spendTransaction(states, txn, totalFee, false)
 		}
+
+		states.CleanupNanoPay(b.Header.UnsignedHeader.Height)
 	}
 
 	var root Uint256
 	if needBeCommitted {
-		root, err = states.CommitTo(true)
+		root, err = states.Finalize(true)
 		if err != nil {
 			return nil, EmptyUint256, err
 		}
 	} else {
-		root = states.IntermediateRoot(true)
+		root = states.IntermediateRoot()
 	}
 
 	return states, root, nil
--- a/chain/db/statedb.go
+++ b/chain/db/statedb.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sort"
 	"sync"
 
 	"github.com/nknorg/nkn/common"
@@ -16,6 +17,12 @@ type Operation int
 const (
 	Addition Operation = iota
 	Subtraction
+)
+
+var (
+	AccountPrefix        = []byte{0x00}
+	NanoPayPrefix        = []byte{0x01}
+	NanoPayCleanupPrefix = []byte{0x02}
 )
 
 type account struct {
@@ -94,9 +101,11 @@ func (acc *account) Empty() bool {
 }
 
 type StateDB struct {
-	db       *cachingDB
-	trie     ITrie
-	accounts sync.Map
+	db             *cachingDB
+	trie           ITrie
+	accounts       sync.Map
+	nanoPay        map[string]*nanoPay
+	nanoPayCleanup map[uint32]map[string]struct{}
 }
 
 func NewStateDB(root common.Uint256, db *cachingDB) (*StateDB, error) {
@@ -105,8 +114,10 @@ func NewStateDB(root common.Uint256, db *cachingDB) (*StateDB, error) {
 		return nil, err
 	}
 	return &StateDB{
-		db:   db,
-		trie: trie,
+		db:             db,
+		trie:           trie,
+		nanoPay:        make(map[string]*nanoPay, 0),
+		nanoPayCleanup: make(map[uint32]map[string]struct{}, 0),
 	}, nil
 }
 
@@ -117,7 +128,7 @@ func (sdb *StateDB) getAccount(addr common.Uint160) (*account, error) {
 		}
 	}
 
-	enc, err := sdb.trie.TryGet(addr[:])
+	enc, err := sdb.trie.TryGet(append(AccountPrefix, addr[:]...))
 	if err != nil || len(enc) == 0 {
 		return nil, fmt.Errorf("[getAccount]can not get account from trie: %v", err)
 	}
@@ -131,6 +142,73 @@ func (sdb *StateDB) getAccount(addr common.Uint160) (*account, error) {
 	sdb.setAccount(addr, data)
 
 	return data, nil
+}
+
+func getNanoPayId(sender, recipient common.Uint160, nonce uint64) string {
+	buf := new(bytes.Buffer)
+	_, _ = sender.Serialize(buf)
+	_, _ = recipient.Serialize(buf)
+	_ = serialization.WriteUint64(buf, nonce)
+	return string(buf.Bytes())
+}
+
+func (sdb *StateDB) getNanoPay(id string) (*nanoPay, error) {
+	var np *nanoPay
+	var ok bool
+	if np, ok = sdb.nanoPay[id]; !ok {
+		enc, err := sdb.trie.TryGet(append(NanoPayPrefix, id...))
+		if err != nil {
+			return nil, err
+		}
+
+		np = &nanoPay{}
+
+		if len(enc) > 0 {
+			buff := bytes.NewBuffer(enc)
+			if err := np.Deserialize(buff); err != nil {
+				return nil, fmt.Errorf("[getNanoPay]Failed to decode state object for nano pay: %v", err)
+			}
+		}
+
+		sdb.nanoPay[id] = np
+	}
+
+	return np, nil
+}
+
+func getNanoPayCleanupId(height uint32) string {
+	buf := new(bytes.Buffer)
+	_ = serialization.WriteUint32(buf, height)
+	return string(buf.Bytes())
+}
+
+func (sdb *StateDB) getNanoPayCleanup(height uint32) (map[string]struct{}, error) {
+	var npc map[string]struct{}
+	var ok bool
+	if npc, ok = sdb.nanoPayCleanup[height]; !ok {
+		enc, err := sdb.trie.TryGet(append(NanoPayCleanupPrefix, getNanoPayCleanupId(height)...))
+		if err != nil || len(enc) == 0 {
+			return nil, fmt.Errorf("[getNanoPayCleanup]can not get nano pay cleanup from trie: %v", err)
+		}
+
+		buff := bytes.NewBuffer(enc)
+		npc = make(map[string]struct{}, 0)
+		npcLength, err := serialization.ReadVarUint(buff, 0)
+		if err != nil {
+			return nil, fmt.Errorf("[getNanoPayCleanup]Failed to decode state object for nano pay cleanup: %v", err)
+		}
+		for i := uint64(0); i < npcLength; i++ {
+			id, err := serialization.ReadVarString(buff)
+			if err != nil {
+				return nil, fmt.Errorf("[getNanoPayCleanup]Failed to decode state object for nano pay cleanup: %v", err)
+			}
+			npc[id] = struct{}{}
+		}
+
+		sdb.nanoPayCleanup[height] = npc
+	}
+
+	return npc, nil
 }
 
 func (sdb *StateDB) GetBalance(addr common.Uint160) common.Fixed64 {
@@ -157,6 +235,17 @@ func (sdb *StateDB) GetID(addr common.Uint160) []byte {
 	}
 
 	return account.GetID()
+}
+
+func (sdb *StateDB) GetNanoPay(sender, recipient common.Uint160, nonce uint64) (common.Fixed64, uint32, error) {
+	id := getNanoPayId(sender, recipient, nonce)
+
+	np, err := sdb.getNanoPay(id)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	return np.balance, np.expiresAt, nil
 }
 
 func (sdb *StateDB) SetAccount(addr common.Uint160, acc *account) {
@@ -189,11 +278,11 @@ func (sdb *StateDB) updateAccount(addr common.Uint160, acc *account) error {
 		panic(fmt.Errorf("can't encode object at %x: %v", addr[:], err))
 	}
 
-	return sdb.trie.TryUpdate(addr[:], buff.Bytes())
+	return sdb.trie.TryUpdate(append(AccountPrefix, addr[:]...), buff.Bytes())
 }
 
-func (sdb *StateDB) deleteAccount(addr common.Uint160, acc *account) error {
-	err := sdb.trie.TryDelete(addr[:])
+func (sdb *StateDB) deleteAccount(addr common.Uint160) error {
+	err := sdb.trie.TryDelete(append(AccountPrefix, addr[:]...))
 	if err != nil {
 		return err
 	}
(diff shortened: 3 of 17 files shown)


chain/db/store.go
```diff
@@ -264,7 +264,7 @@ func (cs *ChainStore) persist(b *Block) error {
 			return err
 		}
 
-		if txn.UnsignedTx.Payload.Type != CoinbaseType && txn.UnsignedTx.Payload.Type != CommitType {
+		if txn.UnsignedTx.Payload.Type != CoinbaseType && txn.UnsignedTx.Payload.Type != CommitType && txn.UnsignedTx.Payload.Type != UnidirectionalPaymentChannelType {
```
gdmmx: How about replace with switch ... case sentence?

gdmmx:
@trueinsider It seems that update channel amount have to wait on-chain consensus for each time?

trueinsider:
Only if you submit it to chain, which you not necessarily should do, you can submit only each 5th tx for example, or only at 5 min intervals.

There is more details: https://forum.nkn.org/t/skipping-nonce-check-on-payment-channel/470/3

gdmmx:
@trueinsider I posted additional comment at
https://forum.nkn.org/t/skipping-nonce-check-on-payment-channel/470/6

yilunzhang:
A (not important) naming suggestion: without lock & prepay, it seems impossible to guarantee the availability of the token being paid as sender can transfer those amount away in advance. So this very convenient but not so safe channel should only be used for micro/nano payment where the risk is very small. So I suggest to change the name to Unidirectional Nano Payment Channel to explicitly indicate the usage.

chain/db/store.go
trueinsider: done

trueinsider:
@yilunzhang @gdmmx implemented suggestions, please look again

chain/db/statedb.go
```diff
@@ -10,12 +10,13 @@ import (
 )
 
 type account struct {
-	nonce   uint64
-	balance common.Fixed64
+	nonce               uint64
+	balance             common.Fixed64
+	nanoPaymentChannels map[string]common.Fixed64
```
NoelBright: Map is a unordered type. So this will make the serialization unequal.

chain/pool/txpool.go
```diff
@@ -175,6 +176,34 @@ func (tp *TxnPool) processTx(txn *Transaction) error {
 			}
 
 		}
+	case pb.NanoPaymentChannelType:
```
NoelBright: txlist sorts transactions by nonce. This NanoPaymentChannelType transaction has no nonce.  Insert transactions with this type will cause an error.

chain/db/statedb.go
trueinsider: you mean if I serialize(1)+deserialize+serialize(2) then byte contents in 1 and 2 will be different? why does it matter though?

chain/pool/txpool.go
trueinsider: every tx has a nonce, because it's on the same level as payload and not optional. it could be 0 or any other value though, since it's not checked, so the sort order would be wrong. what error this will cause?

chain/db/statedb.go
yilunzhang: This is a requirement because the state root is stored in block, so the serialization of world state needs to be bit-wise identical on all nodes, otherwise block proposed by one node may be considered invalid by other nodes because they got a different state root.

trueinsider: ok, fixed

chain/pool/txpool.go
trueinsider: fixed

chain/txValidator.go
```diff
@@ -157,6 +157,28 @@ func CheckTransactionPayload(txn *Transaction) error {
...
+		pld := payload.(*NanoPaymentChannel)
+		if len(pld.ChannelId) > 32 {
+			return errors.New("length of channelId error")
+		}
+
+		if len(pld.Sender) != 20 && len(pld.Recipient) != 20 {
+			return errors.New("length of programhash error")
+		}
+
+		donationProgramhash, _ := ToScriptHash(config.DonationAddress)
```
NoelBright: is the sender need to be donationaddress?

trueinsider: it doesn't, so this check prevents it

NoelBright: how to prevent multiple persons to use this channel id?

trueinsider: why? it should be unique only per receiver

yilunzhang: What about make it unique per sender so that it can be easily extended to multiple receiver?

chain/db/statedb.go
```diff
@@ -2,14 +2,279 @@ package db
...
+	}
+}
+
+func (ga *globalAccount) SetNanoPay(sender, recipient common.Uint160, nonce uint64, balance common.Fixed64, limit uint32) {
+	id := getNanoPayId(sender, recipient, nonce)
+	var np *nanoPay
+	var ok bool
+	if np, ok = ga.nanoPay[id]; !ok {
+		np = &nanoPay{balance: balance, limit: limit}
+		ga.nanoPay[id] = np
```
yilunzhang: Should we call cleanupNanoPayAtHeight() here?

chain/pool/txpool.go
```diff
@@ -87,6 +89,34 @@ func (tp *TxnPool) AppendTxnPool(txn *transaction.Transaction) error {
...
+	amountToClaim := common.Fixed64(0)
+	var err error
+	tp.NanoPayTxs.Range(func(_, v interface{}) bool {
+		tx := v.(*transaction.Transaction)
+		var pld transaction.IPayload
+		pld, err = transaction.Unpack(tx.UnsignedTx.Payload)
+		if err != nil {
+			return false
+		}
+		payloadNp := pld.(*pb.NanoPay)
```
yilunzhang: I think It's better to use `payloadNp, ok := pld.(*pb.NanoPay)` here to avoid panic caused by accidentally put non-nanopay txn into here

chain/txValidator.go
```diff
@@ -271,6 +292,27 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
 		}
+	case pb.NanoPayType:
+		pld := payload.(*pb.NanoPay)
+
+		channelBalance, limit := DefaultLedger.Store.GetNanoPay(
+			BytesToUint160(pld.Sender),
+			BytesToUint160(pld.Recipient),
+			pld.Nonce,
+		)
+
```
yilunzhang: We probably need to check and return error if nanopay does not exist (channelBalance == limit == 0)

chain/pool/txpool.go
```diff
@@ -87,6 +89,34 @@ func (tp *TxnPool) AppendTxnPool(txn *transaction.Transaction) error {
 	return nil
 }
 
+func (tp *TxnPool) totalClaim(sender []byte) (common.Fixed64, error) {
+	amountToClaim := common.Fixed64(0)
+	var err error
+	tp.NanoPayTxs.Range(func(_, v interface{}) bool {
```
yilunzhang: Currently txnpool.NanoPayTxs stores map txnHash -> txn, so for every nanopay txn added, all existing nano txns in txn pool needs to be traversed and the cost is O(N^2) where N is the # of nano pay txn put into txn pool. What about changing it to map sender -> (map txnHash -> txn)? txnpool.TxLists is already doing it now.

util/config/config.go
```diff
@@ -43,6 +43,7 @@ const (
 	DonationAddress              = "NKNaaaaaaaaaaaaaaaaaaaaaaaaaaaeJ6gxa"
 	DonationAdjustDividendFactor = 1
 	DonationAdjustDivisorFactor  = 2
+	GlobalAccountAddress         = "NKNBV71h5oUBjknMHPikPU7nraw9KtksWjtB"
```
yilunzhang: We should also check and make sure no transfer is made to this address

chain/txValidator.go
trueinsider: Why? This could be first claim, so channelBalance + limit would be 0

util/config/config.go
trueinsider: We could, why though? Sending here is effectively burning coins. There is a lot of addresses (public keys) for which nobody will have private key (all zeros, zeros + 1 at the end, etc), we can't possibly check them all.

chain/txValidator.go
yilunzhang: Because `if limit < header.UnsignedHeader.Height` below will always be true for the first claim. You are right about the first claim, but still we need another condition below to make sure the first claim can go through.

util/config/config.go
yilunzhang: I was trying to make sure that no non-leaf node in the the state trie stores account state (though this is probably not necessary at all)

chain/db/statedb.go
```diff
@@ -2,14 +2,279 @@ package db
...
+		if err := serialization.WriteUint32(w, k); err != nil {
+			return fmt.Errorf("globalAccount nanoPayCleanup Serialize error: %v", err)
+		}
+
+		npc := ga.nanoPayCleanup[k]
+		if err := serialization.WriteVarUint(w, uint64(len(npc))); err != nil {
+			return fmt.Errorf("globalAccount nanoPayCleanup Serialize error: %v", err)
+		}
+
+		for id := range npc {
```
NoelBright: this npc is type of 'map[string]struct{}', the keys of it should be out-of-order.

yilunzhang: I think the plan is to change it to trie from map so we don't need this anymore

chain/pool/txpool.go
```diff
@@ -487,6 +572,7 @@ func (tp *TxnPool) verifyTransactionWithLedger(txn *transaction.Transaction) err
 		if len(id) != 0 {
 			return errors.New("ID has be registered")
 		}
+	case pb.NanoPayType:
```
NoelBright: Is there some checks with ledger?

yilunzhang: I don't think so, I think using state is enough (which we should try to achieve for all txn types). But I will wait @trueinsider to confirm.

trueinsider: Actually, I'm confused. In theory there should be exactly the same validation as in txValidator.go. In master they even use exactly the same functions. I'm not sure why it become separate again when we switched to balance.

chain/pool/txpool.go
trueinsider: I understand what you mean, though I'm not sure if we can use txnpool.TxLists as it is (it relies on nonce-ordering and also have concept of orphans, both of which nanopay doesn't have), might need to come up with something else.

chain/pool/txpool.go
yilunzhang: I agree. I think we should do that clean soon

chain/pool/txpool.go
yilunzhang: What I mean is that we don't use the current txn pool (because nanopay txn does not have a nonce), but just change what is stored in NanoPayTxs.

util/config/config.go
yilunzhang: I think we can safely forget about it :)

chain/txValidator.go
trueinsider: resolving, since we moved to global state

chain/db/statedb.go
trueinsider: done

chain/pool/txpool.go
trueinsider: no payload reading in pool anymore

chain/txValidator.go
trueinsider: done

chain/pool/txpool.go
trueinsider: there is no total claim counting in pool anymore

chain/pool/txpool.go
trueinsider: done in separate PR

chain/db/statedb.go
trueinsider: @yilunzhang is right, we won't have that serialization anymore

trueinsider:
@yilunzhang @NoelBright I made changes according to comments, please look again

chain/db/nanopay.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return fmt.Errorf("account nanoPay Serialize error: %v", err)
+	}
+
+	return nil
+}
+
+func (np *nanoPay) Deserialize(r io.Reader) error {
+	var balance common.Fixed64
+	err := balance.Deserialize(r)
```
yilunzhang: Is there a reason not to use `np.balance.Deserialize(r)` here?

chain/db/stateGenerator.go
```diff
@@ -73,6 +73,23 @@ func (cs *ChainStore) spendTransaction(states *StateDB, txn *transaction.Transac
...
 
+	case pb.NanoPayType:
+		nanoPay := pl.(*pb.NanoPay)
+		pg, err := txn.GetProgramHashes()
+		if err != nil {
+			return err
+		}
+
+		addrRecipient := BytesToUint160(nanoPay.Recipient)
+		nanoPayBalance, _ := states.GetNanoPay(pg[0], addrRecipient, nanoPay.Nonce)
```
yilunzhang: I think GetNanoPay should also return an error in addition, otherwise if there is an error, returned balance will be zero, and the claimAmount will be the latest amount

chain/db/statedb.go
```diff
@@ -202,32 +288,98 @@ func (sdb *StateDB) deleteAccount(addr common.Uint160, acc *account) error {
...
+		npcs = append(npcs, id)
+	}
+	sort.Strings(npcs)
+	for _, id := range npcs {
+		if err := serialization.WriteVarString(buff, id); err != nil {
+			panic(fmt.Errorf("can't encode nano pay cleanup %v: %v", nanoPayCleanup, err))
+		}
+	}
+
+	return sdb.trie.TryUpdate(append(NanoPayPrefix, getNanoPayCleanupId(height)...), buff.Bytes())
```
yilunzhang: Should this be NanoPayCleanupPrefix?

chain/db/statedb.go
```diff
@@ -202,32 +288,98 @@ func (sdb *StateDB) deleteAccount(addr common.Uint160, acc *account) error {
...
+}
+
+func (sdb *StateDB) deleteNanoPayCleanup(height uint32) error {
+	err := sdb.trie.TryDelete(append(NanoPayPrefix, getNanoPayCleanupId(height)...))
```
yilunzhang: Should this be NanoPayCleanupPrefix also?

chain/db/statedb.go
```diff
@@ -202,32 +288,98 @@ func (sdb *StateDB) deleteAccount(addr common.Uint160, acc *account) error {
...
 
-func (sdb *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Uint256 {
-	sdb.Finalise(deleteEmptyObjects)
+func (sdb *StateDB) IntermediateRoot() common.Uint256 {
+	sdb.Finalise()
 	return sdb.trie.Hash()
 }
 
-func (sdb *StateDB) CommitTo(deleteEmptyObjects bool) (root common.Uint256, err error) {
+func (sdb *StateDB) CommitTo() (root common.Uint256, err error) {
```
yilunzhang: @NoelBright I think we can merge Finalise() and CommitTo() later (not in this PR), quite a bit redundant code.

chain/pool/txpool.go
```diff
@@ -209,7 +213,13 @@ func (tp *TxnPool) GetTransaction(hash common.Uint256) *transaction.Transaction
 		}
 		return true
 	})
-	return found
+	if found != nil {
+		return found
+	}
+	if np, ok := tp.NanoPayTxs.Load(hash); ok {
```
yilunzhang: I think we should search NanoPayTxs before TxLists because the former one is much faster while TxLists needs to be traversed.

chain/txValidator.go
```diff
@@ -165,15 +165,37 @@ func CheckTransactionPayload(txn *transaction.Transaction) error {
 		if Fixed64(pld.RegistrationFee) < Fixed64(config.MinGenIDRegistrationFee) {
 			return errors.New("fee is too low than MinGenIDRegistrationFee")
 		}
+	case pb.NanoPayType:
+		pld := payload.(*pb.NanoPay)
 
+		if len(pld.Sender) != 20 && len(pld.Recipient) != 20 {
```
yilunzhang: It's better to use UINT160SIZE instead of hard coded 20

yilunzhang: @NoelBright We should also modify the existing ones

chain/txValidator.go
```diff
@@ -421,6 +472,24 @@ func (bvs *BlockValidationState) VerifyTransactionWithBlock(txn *transaction.Tra
 		}
 		bvs.generateIDs[publicKey] = struct{}{}
 		amount = Fixed64(generateIdPayload.RegistrationFee)
+	case pb.NanoPayType:
+		npPayload := payload.(*pb.NanoPay)
+		if npPayload.Height < header.UnsignedHeader.Height {
+			return errors.New("[VerifyTransactionWithBlock], nano pay expired")
+		}
+		key := nanoPay{BytesToHexString(npPayload.Sender), BytesToHexString(npPayload.Recipient), npPayload.Nonce}
```
yilunzhang: why not just using getNanoPayId to compute key?

chain/db/nanopay.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"io"
+
+	"github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/common/serialization"
+)
+
+type nanoPay struct {
+	balance common.Fixed64
+	limit   uint32
```
yilunzhang: What about rename limit -> expiresAt or expiration? When limit occurs together with balance, it sounds like amount limit.

chain/db/statedb.go
```diff
@@ -202,32 +288,98 @@ func (sdb *StateDB) deleteAccount(addr common.Uint160, acc *account) error {
...
+	return nil
+}
+
+func (sdb *StateDB) updateNanoPayCleanup(height uint32, nanoPayCleanup map[string]struct{}) error {
+	buff := bytes.NewBuffer(nil)
+
+	if err := serialization.WriteVarUint(buff, uint64(len(nanoPayCleanup))); err != nil {
+		panic(fmt.Errorf("can't encode nano pay cleanup %v: %v", nanoPayCleanup, err))
+	}
+	npcs := make([]string, 0)
```
yilunzhang: It's more efficient to use `make([]string, len(nanoPayCleanup))` here since map length is stored in Go

chain/db/nanopay.go
trueinsider: I guess not, fixed

chain/db/stateGenerator.go
trueinsider: done

chain/db/statedb.go
trueinsider: thanks, fixed

chain/db/statedb.go
trueinsider: fixed

chain/db/statedb.go
trueinsider: merged

chain/pool/txpool.go
trueinsider: done

chain/txValidator.go
trueinsider: modified wherever found

chain/db/nanopay.go
trueinsider: done

chain/db/statedb.go
trueinsider: fixed

chain/txValidator.go
trueinsider: I guess because current key doesn't need to serialize anything, so it should be faster.

yilunzhang: Haha, then what about switching to getNanoPayId to make it less likely to have bugs?

trueinsider: I'm not sure they're equivalent, e.g. if nano pay state were to be moved into sender's state (for whatever reason), then getNanoPayId wouldn't use sender, but key in validator still must use it.

yilunzhang: I see, that makes sense

NoelBright:
one problem with VerifyTransactionWithBlock.
that is: the modification with BlockValidationState is not atomic in VerifyTransactionWithBlock. It's Ok in verify a block, because a new BlockValidationState is created every time. But in the transaction pool the BlockValidationState is existing as long as the life cycle of nknd. A invalid transaction maybe modify BlockValidationState of txpool in a wrong way.

pb/payload.proto
```diff
@@ -71,3 +72,12 @@ message GenerateID {
 	bytes public_key       = 1;
 	int64 registration_fee = 2;
 }
+
+message NanoPay {
+	bytes sender    = 1;
+    bytes recipient = 2;
```
yilunzhang: The indent seems to be mixed with tab and spaces

chain/pool/txpool.go
```diff
@@ -22,6 +22,7 @@ type TxnPool struct {
 	TxLists              sync.Map // NonceSortedTxs instance to store user's account.
 	TxMap                sync.Map
 	TxShortHashMap       sync.Map
+	NanoPayTxs           sync.Map // tx with nano pay type.
```
NoelBright: How can miner package nanoPay transaction in mining.go? this map is not used in GetAllTransactionLists.

yilunzhang: I think you are right, GetAllTransactionLists() needs to be modified also I believe, though I'm not sure what would be a good order, like regular txn first or nanopay first

trueinsider: I believe it got lost after one of rebases, I'll bring it back.

trueinsider: I believe it got lost after one of rebases, I'll bring it back.

pb/payload.proto
trueinsider: done

chain/pool/txpool.go
trueinsider: fixed

NoelBright:
looks good to be. 

hm, by the way, CleanSubmittedTransactions in txvalidator.go need to be dealt with according to VerifyTransactionWithBlock.

trueinsider:
Even before `CleanSubmittedTransactions` in txValidator is run, txs is already cleared from pool. So, in case of error it'll be invalid either way. Best we could do is to regenerate BlockValidationState from scratch using txs that currently in the pool.

NoelBright:
Looks good to me.

PR #500, opened 2019-05-09
modify wallet store format and modify reward adjust interval to 3 days

NoelBright:
1. modify reward adjust interval to 3 days
2. add nonce to Coinbase transaction
3. modify wallet store format
4. modify API: getnoncebyaddr
5. Compatible with old wallet

### Checklist

### Extra information

### DIFF
--- a/block/block.go
+++ b/block/block.go
@@ -194,7 +194,8 @@ func GenesisBlockInit() (*Block, error) {
 	}
 
 	rewardAddress, _ := ToScriptHash(config.InitialIssueAddress)
-	payload := NewCoinbase(EmptyUint160, rewardAddress, Fixed64(config.InitialIssueAmount))
+	donationProgramhash, _ := ToScriptHash(config.DonationAddress)
+	payload := NewCoinbase(donationProgramhash, rewardAddress, Fixed64(config.InitialIssueAmount))
 	pl, err := Pack(CoinbaseType, payload)
 	if err != nil {
 		return nil, err
--- a/chain/db/store.go
+++ b/chain/db/store.go
@@ -264,12 +264,12 @@ func (cs *ChainStore) persist(b *Block) error {
 			return err
 		}
 
-		if txn.UnsignedTx.Payload.Type != CoinbaseType && txn.UnsignedTx.Payload.Type != CommitType {
-			pg, _ := ToCodeHash(txn.Programs[0].Code)
-			acc := states.GetOrNewAccount(pg)
+		if txn.UnsignedTx.Payload.Type != CommitType {
+			pg, _ := txn.GetProgramHashes()
+			acc := states.GetOrNewAccount(pg[0])
 			nonce := acc.GetNonce()
 			acc.SetNonce(nonce + 1)
-			states.SetAccount(pg, acc)
+			states.SetAccount(pg[0], acc)
 		}
 
 		pl, err := Unpack(txn.UnsignedTx.Payload)
@@ -344,7 +344,6 @@ func (cs *ChainStore) persist(b *Block) error {
 
 	headerRoot, _ := Uint256ParseFromBytes(b.Header.UnsignedHeader.StateRoot)
 	if ok := root.CompareTo(headerRoot); ok != 0 {
-		log.Error("dump trimed block:", headerBuffer)
 		return fmt.Errorf("state root not equal:%v, %v", root.ToHexString(), headerRoot.ToHexString())
 	}
 
--- a/chain/pool/txpool.go
+++ b/chain/pool/txpool.go
@@ -301,6 +301,21 @@ func (tp *TxnPool) GetTxnByCount(num int) (map[common.Uint256]*Transaction, erro
 	return txmap, nil
 }
 
+func (tp *TxnPool) GetNonceByTxnPool(addr common.Uint160) (uint64, error) {
+	if _, ok := tp.TxLists[addr]; !ok {
+		return 0, errors.New("no transactions in transaction pool")
+	}
+
+	list := tp.TxLists[addr]
+	pendingNonce, err := list.GetLatestNonce()
+	if err != nil {
+		return 0, err
+	}
+	expectedNonce := pendingNonce + 1
+
+	return expectedNonce, nil
+}
+
 func (tp *TxnPool) Dump() {
 	tp.mu.RLock()
 	defer tp.mu.RUnlock()
--- a/vault/store.go
+++ b/vault/store.go
@@ -25,10 +25,10 @@ type HeaderData struct {
 }
 
 type AccountData struct {
-	Address             string
-	ProgramHash         string
-	PrivateKeyEncrypted string
-	ContractData        string
+	Address       string
+	ProgramHash   string
+	SeedEncrypted string
+	ContractData  string
 }
 
 type WalletData struct {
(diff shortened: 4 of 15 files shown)


block/block.go
```diff
@@ -194,7 +194,8 @@ func GenesisBlockInit() (*Block, error) {
 	}
 
 	rewardAddress, _ := ToScriptHash(config.InitialIssueAddress)
-	payload := NewCoinbase(EmptyUint160, rewardAddress, Fixed64(config.InitialIssueAmount))
+	donationProgramhash, _ := ToScriptHash(config.DonationAddress)
+	payload := NewCoinbase(donationProgramhash, rewardAddress, Fixed64(config.InitialIssueAmount))
```
yilunzhang: What's the difference between using empty vs donation address as sender?

chain/db/store.go
```diff
@@ -264,12 +264,12 @@ func (cs *ChainStore) persist(b *Block) error {
 			return err
 		}
 
-		if txn.UnsignedTx.Payload.Type != CoinbaseType && txn.UnsignedTx.Payload.Type != CommitType {
-			pg, _ := ToCodeHash(txn.Programs[0].Code)
-			acc := states.GetOrNewAccount(pg)
+		if txn.UnsignedTx.Payload.Type != CommitType {
+			pg, _ := txn.GetProgramHashes()
+			acc := states.GetOrNewAccount(pg[0])
```
yilunzhang: Can we guarantee len(pg) > 0?

chain/db/store.go
```diff
@@ -264,12 +264,12 @@ func (cs *ChainStore) persist(b *Block) error {
 			return err
 		}
 
-		if txn.UnsignedTx.Payload.Type != CoinbaseType && txn.UnsignedTx.Payload.Type != CommitType {
-			pg, _ := ToCodeHash(txn.Programs[0].Code)
-			acc := states.GetOrNewAccount(pg)
+		if txn.UnsignedTx.Payload.Type != CommitType {
+			pg, _ := txn.GetProgramHashes()
```
yilunzhang: It's better to handle error (not just here but everywhere)

chain/pool/txpool.go
```diff
@@ -301,6 +301,21 @@ func (tp *TxnPool) GetTxnByCount(num int) (map[common.Uint256]*Transaction, erro
 	return txmap, nil
 }
 
+func (tp *TxnPool) GetNonceByTxnPool(addr common.Uint160) (uint64, error) {
+	if _, ok := tp.TxLists[addr]; !ok {
+		return 0, errors.New("no transactions in transaction pool")
```
yilunzhang: What about change it to `return 0, nil` because it's not an error, and most of the times this condition will be true.
Also after calling GetNonceByTxnPool and get returns in getNonceByAddr, the condition needs to be changed accordingly.

vault/store.go
```diff
@@ -74,8 +75,18 @@ func LoadStore(fullPath string) (*WalletStore, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	js, err := simplejson.NewJson(fileData)
+	value, ok := js.CheckGet("PrivateKeyEncrypted")
+	if ok {
+		privateKey := value.MustString()
+		js.Set("SeedEncrypted", privateKey[:64])
```
yilunzhang: What about using ed25519.SeedSize * 2? But no need to change if this will be removed later.

vault/store.go
```diff
@@ -10,6 +10,7 @@ import (
 	"runtime"
 	"sync"
 
+	simplejson "github.com/bitly/go-simplejson"
 	. "github.com/nknorg/nkn/common"
 )
```
yilunzhang: Shall we change WalletStoreVersion from 0.0.1 to something like 0.0.2 or 0.1?

block/block.go
NoelBright: this is a bug， only the coinbase transaction of genesisblock has a  EmptyUint160 sender.

chain/db/store.go
NoelBright: pg is exist certainly, because this block has passed block verification.

vault/store.go
NoelBright: Yes, it will be remove later.

vault/store.go
NoelBright: Yes, it needs to be changed synchronously to SDK.

yilunzhang: We can make the change now and people will make changes on the SDK side. It doesn't need to be synchronized because the version is not used anywhere AFAIK.

chain/pool/txpool.go
NoelBright: most times this condition will be false. because if there is no transaction in txpool, this return nonce from this function is invalid.

vault/store.go
NoelBright: I had added in vault/wallet.go.

chain/pool/txpool.go
yilunzhang: Oh actually I also meant most times this condition will be false. This will cause an warning error for every GetNonceByTxnPool call, which I consider as a false alarm because there is nothing wrong with it. The reasons I suggest changing it to `return 0, nil` are:
1. Conceptually this is not an error, it's just the return should be "empty"
2. 0 is the empty value here and should never be returned if tp.TxLists[addr] is not empty.
3. Those warning logs will be gone

NoelBright: The return error can help to locate where return 0, because there are 2 positions to return 0. I can remove the log.warning in getNonceByAddr.

chain/db/store.go
NoelBright: I will handle all these errors everywhere  in next commit .

chain/pool/txpool.go
yilunzhang: If we remove log.warning, then if there is really an error returned by list.GetLatestNonce(), it will not be logged and we'd never know. What I suggest is that, we keep the error log if err is not nil, and we ignore the return value if returned value is zero OR error is not nil, then we can distinguish these two cases. Something like
```suggestion
txPoolNonce, err := txpool.GetNonceByTxnPool(pg)
if txPoolNonce == 0 {
	txPoolNonce = persistNonce
}
if err != nil {
 	log.Errorf("GetNonceByTxnPool error: %v", err)
 }
 if 
```

NoelBright: the other error is  no transaction error too.

yilunzhang: OK I see

chain/db/store.go
```diff
@@ -280,46 +277,80 @@ func (cs *ChainStore) persist(b *Block) error {
...
+			pg, err := txn.GetProgramHashes()
+			if err != nil {
+				return err
+			}
+			accSender := states.GetOrNewAccount(pg[0])
+			amountSender := accSender.GetBalance()
+			accSender.SetBalance(amountSender - Fixed64(txn.UnsignedTx.Fee))
+			nonce := accSender.GetNonce()
+			accSender.SetNonce(nonce + 1)
+			states.SetAccount(pg[0], accSender)
```
yilunzhang: There are quite a lot duplicated codes here. What about creating a function to call these common codes and just call the function here?

NoelBright: em. I do agree with you.

PR #502, opened 2019-05-15
Change sigchain algorithm to Optimize bandwidth and cpu usage

yilunzhang:
* Signature chain now use the hash of msg + VRF of block hash as "signature" and reveal VRF when it's no longer used. Only one VRF computation is required per block epoch, hugely reduces the CPU overhead.
* Sigchain is no longer part of the relay packet. Just the last signature (constant size) is added to the relay packet so that the space overhead is O(1) and < 100 bytes.
* The sigchain will be recovered by backtracking. Each node will only attempt to recover one sigchain per block epoch.

This is a breaking changes.
This has been tested with a local cluster.


pb/sigchain.go
```diff
@@ -361,22 +230,31 @@ func (sc *SigChain) finalSigElem() (*SigChainElem, error) {
 }
 
 func (sc *SigChain) IsFinal() bool {
-	if len(sc.Elems) < 2 || !common.IsEqualBytes(sc.DestPubkey, sc.Elems[len(sc.Elems)-2].NextPubkey) {
+	if sc.Length() < 3 {
 		return false
 	}
-	return true
+	pk := sc.Elems[len(sc.Elems)-2].NextPubkey
+	if pk == nil || bytes.Equal(pk, sc.DestPubkey) {
```
gdmmx: `if pk == nil` is unnecessary because bytes.Equal() had handle it already.
Just `return bytes.Equal(pk, sc.DestPubkey)` directly is enough.

yilunzhang: Actually it's different. What you meant is `pk != nil && ...` while what we have is `pk == nil || ...`.

We allow `pk == nil` here because in order to save space, dest pk is not saved in relay packet, so the last node may not know the receiver client's pk. So we allow the last node's next pk to be nil to indicate that the next hop is the receiver client.

por/porserver.go
```diff
@@ -59,104 +110,111 @@ func GetPorServer() *PorServer {
...
 		return nil, err
 	}
-	sigChain, err := NewSigChainWithSignature(dataSize, dataHash[:], blockHash[:],
-		srcID, srcPubkey, destPubkey, pubKey, signature, proof, sigAlgo, false)
+	sigChain, err := pb.NewSigChainWithSignature(dataSize, blockHash[:], srcID, srcPubkey, destPubkey, pubKey, signature, sigAlgo, false)
 	if err != nil {
 		log.Error("New signature chain with signature error:", err)
 		return nil, err
 	}
+	ps.srcSigChainCache.Add(signature, sigChain)
```
gdmmx: It seems that this `signature` NOT signed by Client, it just an util.RandomBytes(32) from [ws.localNode.SendRelayMessage](https://github.com/nknorg/nkn/pull/502/commits/a5564bc5d7d790eb13be20d1d6344707080a4194#diff-b9add2abd88d41a8ecdb9a966dd818daL39)
Please correct me if I have incorrect understanding.

yilunzhang: You are right. Client side has not implemented signature yet :)

pb/sigchain.go
gdmmx: Oh. Your expectation is `return true` when `pk == nil`.
How about `return pk == nil || bytes.Equal(pk, sc.DestPubkey)`?
`if true then return true` looks verbose.

por/porserver.go
gdmmx: Maybe we should left a `// TODO` comments at here.

pb/sigchain.go
yilunzhang: Agree, just changed it

por/porserver.go
yilunzhang: Don't worry, it will be done in the next few weeks (if not days) 🤣

PR #505, opened 2019-05-21
Prevent sigchain being added to txn pool and allow it to be included in history blocks

yilunzhang:
1. Prevent sigchain txn to be added to txn pool
2. Allow mining sigchain to be included in history blocks

### DIFF
--- a/chain/blockValidator.go
+++ b/chain/blockValidator.go
@@ -284,40 +284,37 @@ func TimestampCheck(timestamp int64) error {
 }
 
 func NextBlockProposerCheck(block *Block) error {
-	winnerHash, winnerType, err := GetNextMiningSigChainTxnHash(block.Header.UnsignedHeader.Height)
+	expectedWinnerHash, expectedWinnerType, err := GetNextMiningSigChainTxnHash(block.Header.UnsignedHeader.Height)
 	if err != nil {
 		return err
 	}
 
-	bwhash, _ := Uint256ParseFromBytes(block.Header.UnsignedHeader.WinnerHash)
-	if winnerHash == EmptyUint256 && bwhash != EmptyUint256 {
-		for _, txn := range block.Transactions {
-			if txn.Hash() == bwhash {
-				_, err = por.NewPorPackage(txn)
-				return err
-			}
-		}
-		return fmt.Errorf("mining sigchain txn %s not found in block", bwhash)
+	winnerType := block.Header.UnsignedHeader.WinnerType
+	winnerHash, err := Uint256ParseFromBytes(block.Header.UnsignedHeader.WinnerHash)
+	if err != nil {
+		return err
 	}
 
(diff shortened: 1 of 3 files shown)


chain/blockValidator.go
```diff
@@ -284,40 +284,37 @@ func TimestampCheck(timestamp int64) error {
...
 		for _, txn := range block.Transactions {
-			if txn.Hash() == bwhash {
+			if txn.Hash() == winnerHash {
 				found = true
 				break
 			}
 		}
 		if !found {
-			return fmt.Errorf("mining sigchain txn %s not found in block", block.Header.UnsignedHeader.WinnerHash)
+			if _, err := DefaultLedger.Store.GetTransaction(winnerHash); err != nil {
```
gdmmx: We should not allow pack winnerHash out of `currentHeight - 1`.
Next proposer should be specific assigned  by current Block. And each block should pack only one sigChain in transactions list.

chain/mining.go
```diff
@@ -47,14 +47,16 @@ func (bm *BuiltinMining) BuildBlock(height uint32, chordID []byte, winningHash c
 	txCount := 1
 
 	if winnerType == TXN_SIGNER {
-		miningSigChainTxn, err := por.GetPorServer().GetMiningSigChainTxn(winningHash)
-		if err != nil {
-			return nil, err
+		if _, err = DefaultLedger.Store.GetTransaction(winnerHash); err != nil {
```
gdmmx: What happen if the winnerHash was in ledger already?
It seems that it will build a block without any sigChains in this proposal.

chain/blockValidator.go
yilunzhang: I'm not sure if it's a good idea.
A sigchain txn is just a normal txn. Any txn, including sigchain txn, could potentially has a >0 fee. It's up to block producer which txn he want to pack. A block proposer has no motivation to pack more (zero fee) sigchain txn in a block because it only reduces the probability that the block will be accepted (e.g. due to not fully propagated).

chain/mining.go
yilunzhang: Yes, but it will not cause any problem.

chain/pool/txpool.go
```diff
@@ -334,17 +336,6 @@ func (tp *TxnPool) verifyTransactionWithLedger(txn *Transaction) error {
 		return ErrDuplicatedTx
 	}
 
-	// get signature chain from commit transaction then add it to POR server
-	if txn.UnsignedTx.Payload.Type == pb.CommitType {
```
gdmmx: Move the TypeCheck into switch ... case maybe cause many unnecessary `Unpack(txn.UnsignedTx.Payload)`

yilunzhang: The Unpack will be removed later when we merge some part of the verification with ledger

gdmmx:
LGTM

PR #506, opened 2019-05-22
V0.9: improve mining and transaction pool

NoelBright:
1. improve mining and transaction pool
2. modify the block verification for new packaging method
3. add processing flow with fee 

### DIFF
--- a/chain/db/store.go
+++ b/chain/db/store.go
@@ -253,6 +253,11 @@ func (cs *ChainStore) persist(b *Block) error {
 		return err
 	}
 
+	var totalFee Fixed64
+	for _, txn := range b.Transactions {
+		totalFee += Fixed64(txn.UnsignedTx.Fee)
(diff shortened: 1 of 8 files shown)


chain/db/store.go
```diff
@@ -280,46 +277,80 @@ func (cs *ChainStore) persist(b *Block) error {
...
+			nonce := accSender.GetNonce()
+			accSender.SetNonce(nonce + 1)
+			states.SetAccount(pg[0], accSender)
 			err = cs.DeleteName(deleteNamePayload.Registrant)
 			if err != nil {
 				return err
 			}
 		case SubscribeType:
 			subscribePayload := pl.(*Subscribe)
+			pg, err := txn.GetProgramHashes()
```
yilunzhang: We will need to move these repeated code to a common function later.

PR #509, opened 2019-05-24
V0.9: add GenerateID function

NoelBright:
add GenerateID function
1. add GenerateID transaction in pb/payload.proto.
2. add GenerateID processing flow in statedb and store.
3. add verification in txpool, mining and tranactioncheck.
4. add API 'getid' and nknc command 'id' .


api/common/helper.go
```diff
@@ -110,3 +110,25 @@ func MakeSubscribeTransaction(wallet vault.Wallet, identifier string, topic stri
 
 	return txn, nil
 }
+
+func MakeGenerateIDTransaction(wallet vault.Wallet, regFee Fixed64, nonce uint64, fee Fixed64) (*Transaction, error) {
```
yilunzhang: what about renaming `fee` as 'txFee` here and also in NewGenerateIDTransaction to make the difference between these 2 fee more explict?

api/common/interfaces.go
```diff
@@ -491,6 +491,40 @@ func getNonceByAddr(s Serverer, params map[string]interface{}) map[string]interf
...
+		return respPacking(nil, INVALID_PARAMS)
+	}
+
+	id, err := chain.DefaultLedger.Store.GetID(pkSlice)
+	if err != nil {
+		return respPacking(nil, INVALID_PARAMS)
+	}
+
+	if len(id) == 0 || bytes.Equal(id, make([]byte, 32)) {
+		return respPacking(nil, INTERNAL_ERROR)
```
yilunzhang: We need to have a special error code if pk does not have an id yet so that the node knows he needs to generate an ID

chain/db/statedb.go
```diff
@@ -61,8 +81,12 @@ func (acc *account) SetBalance(balance common.Fixed64) {
 	acc.balance = balance
 }
 
+func (acc *account) SetID(id []byte) {
+	acc.id = id
+}
+
 func (acc *account) Empty() bool {
-	return acc.nonce == 0 && acc.balance == 0
+	return acc.nonce == 0 && acc.balance == 0 && acc.id == nil
```
yilunzhang: should acc.id be nil or sha256zerohash?

chain/mining.go
```diff
@@ -108,16 +108,17 @@ func (bm *BuiltinMining) BuildBlock(height uint32, chordID []byte, winnerHash co
...
+	curStateHash := GenerateStateRoot(txnList, height, consensusData)
 	header := &Header{
 		BlockHeader: BlockHeader{
 			UnsignedHeader: &UnsignedHeader{
 				Version:          HeaderVersion,
 				PrevBlockHash:    curBlockHash.ToArray(),
 				Timestamp:        timestamp,
 				Height:           height,
-				ConsensusData:    rand.Uint64(),
+				ConsensusData:    consensusData,
```
yilunzhang: My suggestion:
1. Rename ConsensusData to RandomBeacon
2. Change it to 64 bytes (to provide enough entropy)
3. Do not use math.rand (it's not cryptographically secure), use RandomBytes defined in crypto/util/util.go

chain/stateGenerator.go
```diff
@@ -62,16 +66,62 @@ func spendTransaction(states *db.StateDB, tx *Transaction, totalFee common.Fixed
...
+		donationAccount.SetBalance(amount + common.Fixed64(genID.RegistrationFee))
+		states.SetAccount(donationAddress, donationAccount)
+
 	}
 
 	return nil
 }
 
-func GenerateStateRoot(txs []*Transaction) common.Uint256 {
+func GenerateStateRoot(txs []*Transaction, height uint32, consensusData uint64) common.Uint256 {
```
yilunzhang: GenerateStateRoot() and spendTransaction() share lots of code with persist(). What about put them in the same function?

chain/db/statedb.go
```diff
@@ -38,11 +43,22 @@ func (acc *account) Deserialize(r io.Reader) error {
 
 	nonce, err := serialization.ReadVarUint(r, 0)
 	if err != nil {
-		return fmt.Errorf("Deserialize error:%v", err)
+		return fmt.Errorf("Deserialize nonce error:%v", err)
```
yilunzhang: We need to modify this part at the beginning of the function.
```
	if acc == nil {
		acc = new(account)
	}
```

chain/stateGenerator.go
NoelBright: I will optimize it later.

chain/mining.go
gdmmx: @yilunzhang  Just only **random from proposer** for ConsensusData is not enough. 

As an ID register who aim to gain appointed chord position, its pubKey and expected ID is determinate. The hash of generateID txn is also determinate too.

The position picker can exhaust applicable `ConsensusData` **in advance** and waiting for its conspirator win a proposal right.

So, We should add `additional constraint`(**factor from out of current proposer**) for ConsensusData/RandomBeacon generation.


> 2. Change it to 64 bytes to provide enough entropy for most purposes

Promote Uint64 to 64Bytes is expensive for size of Blk.Header.
In v0.9, the average of Blk.Size with only 1 transaction(coinbase) is around 390+ bytes.  +56 bytes means that Blk.Header increase 15% size.

yilunzhang: For 1: You are right, I forget this one. We need to use
id = hash(current block random beacon, prev block hash, txn hash)
@NoelBright 

For 2: I didn't realize our current block size is so small. What about 32 bytes then?

NoelBright: ok

gdmmx:
Looks Good To Me

chain/txValidator.go
```diff
@@ -9,11 +9,13 @@ import (
 
 	"github.com/nknorg/nkn/block"
 	. "github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/crypto"
 	. "github.com/nknorg/nkn/pb"
 	. "github.com/nknorg/nkn/transaction"
 	"github.com/nknorg/nkn/util/address"
 	"github.com/nknorg/nkn/util/config"
 	"github.com/nknorg/nkn/vm/signature"
+	"github.com/ontology/common"
```
trueinsider: I think this is probably a typo

yilunzhang: I believe this is generated automatically by goimports. Every time I use log.xxx it will import nnet/log for me 😂

PR #510, opened 2019-05-25
Start block verification after sigchain cached is filled

yilunzhang:
This prevent new node from blocking consensus due to empty sigchain cache.


consensus/consensus.go
```diff
@@ -280,7 +274,7 @@ func (consensus *Consensus) saveAcceptedBlock(electedBlockHash common.Uint256) e
 
 	log.Infof("Accepted block height: %d, local ledger block height: %d, sync needed.", block.Header.UnsignedHeader.Height, chain.DefaultLedger.Store.GetHeight())
 
-	elc, loaded, err := consensus.loadOrCreateElection(heightToKey(block.Header.UnsignedHeader.Height))
+	elc, loaded, err := consensus.loadOrCreateElection(block.Header.UnsignedHeader.Height)
```
gdmmx: Why changed map's key from hash to height?
We should consider about elected in forked consensus, proposals from different blockChain maybe have same height.

yilunzhang: The key is not changed, heightToKey will be called inside function loadOrCreateElection

consensus/state.go
```diff
@@ -31,6 +40,7 @@ func (consensus *Consensus) startGettingNeighborConsensusState() {
 			if localConsensusHeight == 0 || localConsensusHeight+1 < majorityConsensusHeight {
 				if majorityConsensusHeight+1 > localLedgerHeight {
 					consensus.setNextConsensusHeight(majorityConsensusHeight + 1)
+					consensus.localNode.SetMinVerifiableHeight(majorityConsensusHeight + 1 + por.SigChainMiningHeightOffset)
```
gdmmx: Could we make sure init node.minVerifiableHeight at here earlier than voting start?
As node.minVerifiableHeight uninitialized, it just only `0 + 1 + SigChainMiningHeightOffset = 5`

yilunzhang: It is initialized at the beginning of startGettingNeighborConsensusState()

gdmmx:
LGTM

PR #511, opened 2019-05-30
Use absolute weight instead of relative weight for voting threshold

yilunzhang:
Previously we changed from absolute weight to relative weight for voting success threshold in order to solve the problem of many node restart. Later on when we introduced min verifiable height to solve this problem more robustly, we didn't change the threshold back from relative weight to absolute weight, which may cause the forked region to continue producing blocks.


yilunzhang:
Will be included in another PR

PR #512, opened 2019-05-30
Implement node to node authenticated encryption

yilunzhang:
* Implement hop by hop encryption for all node to node traffic
* Change all msg type to unsigned (since authenticated encryption provides the same security guarantee)
* Verify node verison/id/port before dialing
* Remove wait after block persist due to block persist API change

Note: this is a completely breaking change. New version and old version are speaking different language :)


crypto/ed25519/ed25519.go
```diff
@@ -7,10 +7,15 @@ import (
 
 	"github.com/nknorg/nkn/crypto/ed25519/vrf"
 	"github.com/nknorg/nkn/crypto/util"
+	"github.com/nknorg/vxeddsa/extra25519"
 
 	"golang.org/x/crypto/ed25519"
 )
 
+const PublicKeySize = ed25519.PublicKeySize
```
gdmmx: I am afraid that make confuse `ed25519.PublicKeySize` between "golang.org/x/crypto/ed25519" and "gitHub.com/nknorg/nkn/crypto/ed25519"

yilunzhang: We should be fine as long as we don't confuse the compiler :)

PS: I think this will be removed soon when we drop p256 support

PR #514, opened 2019-06-03
Efficient block and sigchain broadcasting

yilunzhang:
* Implement fast block propagation by only requesting txn hash: Previously a block proposal is transmitted with all txn in it. However, this will lead to significant block progation delay and bandwidth waste if most txnx in the proposal are already progagated enough before proposal is sent. Therefore, it's enough to only include a list of txn short hash or hash in block proposal, and receiver will request the missing txns in it.

* Implement efficient sigchain broadcasting: Previously each node will send sigchain (commit) txn directly to all neighbors, so each node will receive the same sigchain txn once from all its neighbors. Now each node will first send the height and sighash of sigchain txn to its neighbors, and its neighbor will request it if it's the most optimal sigchain so far for the height. This will greatly reduce sigchain broadcasting overhead.

* Add transactions root check


gdmmx:
LGTM

NoelBright:
Looks good to me.

PR #517, opened 2019-06-05
improve state db flow

NoelBright:
move stateGenerator.go to db folder. Use GenerateStateRoot for mining.go
and store.go


chain/db/statedb.go
```diff
@@ -263,3 +271,37 @@ func (sdb *StateDB) SetID(addr common.Uint160, id []byte) error {
...
+		}
+		acc.SetBalance(amount - value)
+	default:
+		return errors.New("UpdateBalance: invalid operation")
+	}
+
+	return nil
+}
+
+func (sdb *StateDB) IncrNonce(addr common.Uint160) error {
```
yilunzhang: Can we guarantee that IncrNonce will not be called concurrently?

chain/db/store.go
```diff
@@ -78,13 +78,23 @@ func (cs *ChainStore) InitLedgerStoreWithGenesisBlock(genesisBlock *block.Block)
...
 
-		root := cs.GetCurrentBlockStateRoot()
-		fmt.Println("---------", root)
-		cs.States, _ = NewStateDB(root, NewTrieStore(cs.GetDatabase()))
+		root, err := cs.GetCurrentBlockStateRoot()
+		if err != nil {
+			return 0, nil
+		}
+
+		fmt.Println("state root:", root.ToHexString())
```
yilunzhang: Why not using log here?

chain/db/statedb.go
NoelBright: account is proccessed sequentially now.

chain/db/store.go
NoelBright: done

chain/db/store.go
```diff
@@ -464,7 +372,10 @@ func (cs *ChainStore) persist(b *block.Block) error {
 		return err
 	}
 
-	cs.States = states
+	cs.States, err = NewStateDB(root, NewTrieStore(cs.GetDatabase()))
```
yilunzhang: If we get error here, the above persist functions are already executed, but cs.States will be empty. Will that cause a problem?

NoelBright: hm, how about panic when returning error?

PR #519, opened 2019-06-05
Optimize block propagation and block building for slow devices

yilunzhang:
* Send i have block msg before verification for faster block propagation
* Add context timeout for building block

### Checklist

### Extra information

### DIFF
--- a/consensus/proposal.go
+++ b/consensus/proposal.go
@@ -144,11 +144,7 @@ func (consensus *Consensus) waitAndHandleProposal() (*election.Election, error) 
 				} else if err = chain.NextBlockProposerCheck(proposal.Header); err != nil {
 					log.Warningf("Proposal fails to pass next block proposal check: %v", err)
 					acceptProposal = false
-				}
-			}
-
-			if acceptProposal {
-				if err = chain.TransactionCheck(ctx, proposal); err != nil {
+				} else if err = chain.TransactionCheck(ctx, proposal); err != nil {
 					log.Warningf("Proposal fails to pass transaction check: %v", err)
 					acceptProposal = false
 				}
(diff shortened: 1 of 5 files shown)


consensus/proposal.go
```diff
@@ -114,9 +114,15 @@ func (consensus *Consensus) waitAndHandleProposal() (*election.Election, error)
 
 			proposals[blockHash] = proposal
 			if len(proposals) > 2 {
-				log.Warningf("Received more than 2 different proposals, ignoring the rest to prevent spam")
+				log.Warningf("Received more than 2 different proposals, ignoring the rest")
 				continue
 			}
+
+			err = consensus.iHaveProposal(consensusHeight, blockHash)
```
gdmmx: propagation before verification will introduce security risk: spam && DDoS(block network) via broadcast huge illegal proposal.

yilunzhang: SignerCheck is still before propagation. Actually you can see that this patch does not change when the block will be propagated: the first (up to) two blocks proposed by the correct proposer.

gdmmx: You are right. Previous implement:
```
if !consensus.canVerifyHeight                  // I can't judge this proposal legal or not
    send iHaveProposal(height, hash)       // propagated continue

acceptProposal = Check timestamp, header, nextWinner, transaction all pass
if acceptProposal
    set myVote
else
    send iHaveProposal(height, hash)       // BUT, WHY still propagated a proposal even I regarded as unacceptable Proposal?
```
**I think the last line code has potential DDoS risk.**

yilunzhang: Because not accepting the block not necessarily means the block is invalid, it may be local node's problem. To prevent DDoS attack, we actually only allow 2 blocks (proposed by the correct proposer) to be propagated. The second one will make sure all nodes will accept neither of the blocks.

PR #522, opened 2019-06-06
Unify and optimize tx validation

trueinsider:
Unify tx validation (remove validation logic duplication).  
Optimize performance for validation of txs in the pool with other txs in the pool.

### Checklist

### DIFF
--- a/chain/pool/txpool.go
+++ b/chain/pool/txpool.go
@@ -1,44 +1,32 @@
 package pool
 
 import (
-	"bytes"
 	"errors"
-	"fmt"
 	"sort"
 	"sync"
 
 	"github.com/nknorg/nkn/chain"
 	"github.com/nknorg/nkn/common"
-	"github.com/nknorg/nkn/crypto"
 	"github.com/nknorg/nkn/pb"
 	"github.com/nknorg/nkn/transaction"
 	"github.com/nknorg/nkn/util/config"
 	"github.com/nknorg/nkn/util/log"
-	"github.com/syndtr/goleveldb/leveldb"
 )
 
 var (
-	ErrDuplicatedTx          = errors.New("duplicate transaction check faild")
-	ErrDoubleSpend           = errors.New("IsDoubleSpend check faild")
-	ErrTxnType               = errors.New("invalidate transaction payload type")
-	ErrNonceTooLow           = errors.New("nonce is too low")
-	ErrDuplicateName         = errors.New("name has be registered already")
-	ErrDuplicateRegistrant   = errors.New("Registrant has already registered this name")
-	ErrNoNameRegistered      = errors.New("name has not be registered yet")
-	ErrDuplicateSubscription = errors.New("Duplicate subscription in one block")
-	ErrDuplicateGenerateID   = errors.New("Duplicate GenerateIDType txn in one block")
-	ErrSubscriptionLimit     = errors.New("Subscription limit exceeded in one block")
+	ErrDuplicatedTx = errors.New("duplicate transaction check failed")
 )
 
 // TxnPool is a list of txns that need to by add to ledger sent by user.
 type TxnPool struct {
-	TxLists        sync.Map // NonceSortedTxs instance to store user's account.
-	TxMap          sync.Map
-	TxShortHashMap sync.Map
+	TxLists              sync.Map // NonceSortedTxs instance to store user's account.
+	TxMap                sync.Map
+	TxShortHashMap       sync.Map
+	blockValidationState *chain.BlockValidationState
 }
 
 func NewTxPool() *TxnPool {
-	return &TxnPool{}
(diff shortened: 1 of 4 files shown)


chain/pool/txpool.go
```diff
@@ -105,161 +98,64 @@ func (tp *TxnPool) processTx(txn *transaction.Transaction) error {
...
 
-	if err := list.AddOrphanTxn(txn); err != nil {
-		return err
+		if replace {
+			if err := list.Add(txn); err != nil {
+				return err
+			}
+		}
+
+		if err := list.Push(txn); err != nil {
```
yilunzhang: should we have an `else` here?

trueinsider: you're right, done

chain/txValidator.go
```diff
@@ -264,107 +285,199 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
 	return nil
 }
 
-type Iterator interface {
-	Iterate(handler func(item *transaction.Transaction) error) error
+type subscription struct {
+	topic      string
+	bucket     uint32
+	subscriber Uint160
```
yilunzhang: I think we should probably change to pk here instead of Uint160 because sender may want to use some of the subscribers' pk for e2e encryption...

yilunzhang: Oh nevermind...

chain/txValidator.go
```diff
@@ -264,107 +285,199 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
+	identifier string
+}
+
+type BlockValidationState struct {
+	txnlist           map[Uint256]struct{}
```
yilunzhang: I think all theses maps needs to be sync.Map because append to txn pool is called concurrently (which calls process txn which calls verify txn with block)

yilunzhang: Also sync.Map LoadOrStore would be quite good for our atomic verification purpose

chain/txValidator.go
```diff
@@ -264,107 +285,203 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
-	type subscription struct{ topic, subscriber string }
-	subscriptions := make(map[subscription]struct{}, 0)
+	switch txn.UnsignedTx.Payload.Type {
+	case pb.CoinbaseType:
+		coinbase := payload.(*pb.Coinbase)
+		donation, err := DefaultLedger.Store.GetDonation()
+		if err != nil {
+			return err
+		}
+		if header != nil && Fixed64(coinbase.Amount) != GetRewardByHeight(header.UnsignedHeader.Height)+donation.Amount {
```
NoelBright: "if header != nil" this judgement can place in front of "coinbase := payload.(*pb.Coinbase)"

trueinsider: done

chain/txValidator.go
trueinsider: done

chain/txValidator.go
```diff
@@ -188,6 +187,9 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
 		return errors.New("Unpack transactiion's paylaod error")
 	}
 
+	sender, _ := ToCodeHash(txn.Programs[0].Code)
```
yilunzhang: Coinbase txn does not have programs, so this will cause panic

yilunzhang: We should also try to avoid omitting error using "_"

trueinsider: done

chain/txValidator.go
```diff
@@ -264,107 +297,210 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
+			delete(bvs.generateIDs, publicKey)
 		}
 
-		return nil
-	})
-
-}
+		if amount > 0 || fee > 0 {
+			if _, ok := bvs.totalAmount[sender]; ok {
+				bvs.totalAmount[sender] -= amount + fee
```
NoelBright: need to check that bvs.totalAmount[sender] > amount + fee?

NoelBright: and if bvs.totalAmount[sender] is not existed, how to deal with it?

yilunzhang: Seems that it's checked in VerifyTransactionWithBlock. This function (CleanSubmittedTransactions) is to rollback BlockValidationState state

yilunzhang: Seems that bvs.totalAmount[sender] should always exist if txns has been applied to BlockValidationState, so we should probably return error if bvs.totalAmount[sender] does not exist or bvs.totalAmount[sender] < amount + fee. @trueinsider

trueinsider: done

NoelBright:
looks good to me.

PR #529, opened 2019-06-10
Add max voting interval and configurable MaxNumTxnPerBlock

yilunzhang:
* Add max voting interval to prevent mixing votes between rounds
* Add MaxNumTxnPerBlock in config to control self proposed block size

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -170,6 +170,8 @@ func (consensus *Consensus) startElection(height uint32, elc *election.Election)
 		return common.EmptyUint256, fmt.Errorf("Convert election result to block hash error")
 	}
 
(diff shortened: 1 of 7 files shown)


consensus/consensus.go
```diff
@@ -44,7 +44,7 @@ type Consensus struct {
 
 // NewConsensus creates a MOCA consensus
 func NewConsensus(account *vault.Account, localNode *node.LocalNode) (*Consensus, error) {
-	txnCollector := chain.NewTxnCollector(localNode.GetTxnPool(), config.MaxNumTxnPerBlock)
+	txnCollector := chain.NewTxnCollector(localNode.GetTxnPool(), int(config.Parameters.MaxNumTxnPerBlock))
```
NoelBright: It is easy to confuse between  config.MaxNumTxnPerBlock with config.Parameters.MaxNumTxnPerBlock

yilunzhang: Agreed, renamed to NumTxnPerBlock

NoelBright:
looks good to me.

PR #533, opened 2019-06-11
use GenerateID transaction to generate chord id

NoelBright:
1. genesisChoidID = zerotxhash + zeroblockhash + block(0).randBeacon
2. when a node try to join nkn network:
2.1 look up local ledger to get chord id; if not found,
2.2 request chord id from remote seed node; if not foud,
2.3 create a chord id by sending a GenerateID transaction.
2.4 request chord id from remote node periodically.
2.4 if time out, quit nknd.

### DIFF
--- a/chain/db/stateGenerator.go
+++ b/chain/db/stateGenerator.go
@@ -7,6 +7,7 @@ import (
 	"github.com/nknorg/nkn/pb"
 	"github.com/nknorg/nkn/transaction"
 	"github.com/nknorg/nkn/util/config"
+	"github.com/nknorg/nkn/vm/contract"
 	"github.com/nknorg/nnet/log"
 )
 
@@ -120,6 +121,23 @@ func (cs *ChainStore) generateStateRoot(b *block.Block, genesisBlockInitialized,
 
 	//process previous block
 	height := b.Header.UnsignedHeader.Height
+	if height == 0 {
+		id := ComputeID(EmptyUint256, EmptyUint256, b.Header.UnsignedHeader.RandomBeacon)
+
+		pk, err := crypto.NewPubKeyFromBytes(b.Header.UnsignedHeader.Signer)
+		if err != nil {
+			return nil, EmptyUint256, err
+		}
+		contract, err := contract.CreateSignatureContract(pk)
+		if err != nil {
+			return nil, EmptyUint256, err
+		}
+
+		if err := states.UpdateID(contract.ProgramHash, id); err != nil {
+			return nil, EmptyUint256, err
+		}
+	}
+
 	if height > config.GenerateIDBlockDelay {
 		prevBlock, err := cs.GetBlockByHeight(height - config.GenerateIDBlockDelay)
 		if err != nil {
@@ -130,17 +148,16 @@ func (cs *ChainStore) generateStateRoot(b *block.Block, genesisBlockInitialized,
 
 		for _, txn := range prevBlock.Transactions {
 			if txn.UnsignedTx.Payload.Type == pb.GenerateIDType {
-				txnHash := txn.Hash()
-				data := append(preBlockHash[:], txnHash[:]...)
-				data = append(data, b.Header.UnsignedHeader.RandomBeacon...)
-				id := crypto.Sha256(data)
+				id := ComputeID(preBlockHash, txn.Hash(), b.Header.UnsignedHeader.RandomBeacon)
 
 				pg, err := txn.GetProgramHashes()
 				if err != nil {
 					return nil, EmptyUint256, err
 				}
 
-				states.UpdateID(pg[0], id)
+				if err := states.UpdateID(pg[0], id); err != nil {
+					return nil, EmptyUint256, err
+				}
 
 			}
 		}
@@ -175,4 +192,11 @@ func (cs *ChainStore) generateStateRoot(b *block.Block, genesisBlockInitialized,
 	}
 
 	return states, root, nil
+}
+
+func ComputeID(preBlockHash, txnHash Uint256, randomBeacon []byte) []byte {
+	data := append(preBlockHash[:], txnHash[:]...)
+	data = append(data, randomBeacon...)
+	id := crypto.Sha256(data)
+	return id
 }
--- a/nknd.go
+++ b/nknd.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"bytes"
+	"encoding/hex"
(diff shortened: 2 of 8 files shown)


chain/db/stateGenerator.go
```diff
@@ -120,6 +121,25 @@ func (cs *ChainStore) generateStateRoot(b *block.Block, genesisBlockInitialized,
 
 	//process previous block
 	height := b.Header.UnsignedHeader.Height
+	if height == 0 {
+		preBlockHash := EmptyUint256
```
yilunzhang: What about creating a common function like ComputeID(preBlockHash, txnHash, randomBeacon) to share between here and below (around line 153)?

chain/db/stateGenerator.go
```diff
@@ -120,6 +121,25 @@ func (cs *ChainStore) generateStateRoot(b *block.Block, genesisBlockInitialized,
...
+		pk, err := crypto.NewPubKeyFromBytes(b.Header.UnsignedHeader.Signer)
+		if err != nil {
+			return nil, EmptyUint256, err
+		}
+		contract, err := contract.CreateSignatureContract(pk)
+		if err != nil {
+			return nil, EmptyUint256, err
+		}
+
+		states.UpdateID(contract.ProgramHash, id)
```
yilunzhang: We should either handle error here (and below), or remove the error returned by UpdateID (because it always returns nil now)

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func GetChordID(seeds []string, publickey []byte) ([]byte, error) {
+	id, err := chain.DefaultLedger.Store.GetID(publickey)
+	if err == nil {
+		if len(id) != 0 && !bytes.Equal(id, make([]byte, 32)) {
```
yilunzhang: We can define a constant instead of hard code 32 here

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
...
+			return id, nil
+		}
+	}
+
+	log.Warningf("get chord ID from localnode faild, no chord ID in ledger")
```
yilunzhang: I think this should be Info instead of warning because it's expected when node first join

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func GetChordID(seeds []string, publickey []byte) ([]byte, error) {
+	id, err := chain.DefaultLedger.Store.GetID(publickey)
+	if err == nil {
```
yilunzhang: We should probably log error here because even if node join for the first time, the expected err should still be nil

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
...
+	log.Warningf("get chord ID from localnode faild, no chord ID in ledger")
+
+	rand.Shuffle(len(seeds), func(i int, j int) {
+		seeds[i], seeds[j] = seeds[j], seeds[i]
+	})
+
+	for _, seed := range seeds {
+		id, err := client.GetChordID(seed, publickey)
+		if err == nil {
+			return id, nil
```
yilunzhang: we should also check length here just in case the seed node is broken or malicious

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
...
+		return nil, err
+	}
+
+	timeout := time.After(time.Minute * 3)
+out:
+	for {
+		select {
+		case <-timeout:
+			break out
+		case <-time.After(time.Second * 40):
```
yilunzhang: Instead of hard code time here, we should wait 2 * config.ConsensusDuration initially, and then check every config.ConsensusDuration

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
...
+		return id, nil
+	}
+
+	log.Warningf("chord ID is not found in local db or remote node: %v", err)
+
+	if err := CreateChordID(seeds, wallet, regFee); err != nil {
+		return nil, err
+	}
+
+	timeout := time.After(time.Minute * 3)
```
yilunzhang: We should use 3 * config.ConsensusTimeout (or even 5 * ) instead of hard code duration

nknd.go
```diff
@@ -375,3 +384,112 @@ func main() {
...
+	account, err := wallet.GetDefaultAccount()
+	if err != nil {
+		return nil, err
+	}
+	pkEncoded, err := account.PubKey().EncodePoint(true)
+	if err != nil {
+		return nil, err
+	}
+
+	id, err := GetChordID(seeds, pkEncoded)
```
yilunzhang: We should try to differentiate 2 cases here:
1. node has sent generate id txn, but id has not been generated yet (id is 32 bytes of zero)
2. node has not generated id yet (id is nil)

node/localnode.go
```diff
@@ -196,15 +195,6 @@ func (localNode *LocalNode) shouldConnectToNode(n *nnetpb.Node) error {
 		}
 	}
 
-	addr, err := url.Parse(n.GetAddr())
-	if err != nil {
-		return err
-	}
-
-	if n.GetId() != nil && !bytes.Equal(address.GenChordID(addr.Host), n.GetId()) {
```
yilunzhang: We should add verification logic here:
1. try to get remote node id (using nodeData.PublicKey)
2. if get id success, compare it with n.GetId(), return err if not match
3. if get id failed, check if local node is persist finished, return err if local node is persist finished

yilunzhang:
A naming suggestion: we can rename createChordId, getChordId, etc to just createId, getId, etc because it has nothing to do with chord.

chain/db/stateGenerator.go
NoelBright: good idea.

nknd.go
NoelBright: it has be handled in GetChordID

node/localnode.go
NoelBright: if getid failed and local node is not persist finished, how to deal with it?

yilunzhang: then we should ignore it and just return nil

nknd.go
yilunzhang: I mean here if we find that node has sent generate id txn, but just has not get an id yet, it should not send new generate id txn or return, but should start wait and get id from rpc right away (line 482(

NoelBright: Yes, I'am fix it.

PR #534, opened 2019-06-12
Prevent inconsistent timestamp validation result on boundary

yilunzhang:
* Change timestamp check to one direction (future only)
* Exclude last rejected proposers
* Check txn root for REQUEST_FULL_TRANSACTION response


yilunzhang:
Will move to #538 to solve the same problem

PR #543, opened 2019-06-17
remove orphan txn list from txpool

NoelBright:
remove orphan txn list from txpool, the txn witch has continuous
nonce can be append to txn pool, otherwise reject it.

### DIFF
--- a/chain/pool/txpool.go
+++ b/chain/pool/txpool.go
@@ -45,20 +45,6 @@ func (tp *TxnPool) AppendTxnPool(txn *transaction.Transaction) error {
 		return errors.New("txpool full, too many transaction in txpool")
 	}
 
-	// 1. process all Orphens
-	tp.TxLists.Range(func(_, v interface{}) bool {
-		if list, ok := v.(*NonceSortedTxs); ok {
-			list.ProcessOrphans(tp.processTx)
-		}
-		return true
-	})
-	tp.TxLists.Range(func(_, v interface{}) bool {
-		if list, ok := v.(*NonceSortedTxs); ok {
-			list.CleanOrphans(nil)
-		}
-		return true
-	})
-
 	// 2. verify txn
 	if err := chain.VerifyTransaction(txn); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


chain/pool/txpool.go
```diff
@@ -138,50 +110,44 @@ func (tp *TxnPool) processTx(txn *transaction.Transaction) error {
...
-				//TODO need more fee
-				isOrphan = false
-				replace = true
-			} else if list.Full() {
-				return errors.New("txpool is full")
-			} else if preNonce, _ := list.GetLatestNonce(); txn.UnsignedTx.Nonce == preNonce+1 {
-				isOrphan = false
+		if _, err := list.Get(txn.UnsignedTx.Nonce); err == nil {
+			log.Warning("replace old tx")
+			if err := tp.blockValidationState.VerifyTransactionWithBlock(txn, nil); err != nil {
```
yilunzhang: Should we replace the old txn in blockValidationState instead of trying to add it to blockValidationState? also @trueinsider

trueinsider: yeah, we should clean old tx before adding new one

chain/pool/txpool.go
```diff
@@ -195,8 +161,9 @@ func (tp *TxnPool) GetAddressList() map[common.Uint160]int {
 	tp.TxLists.Range(func(k, v interface{}) bool {
 		if programHash, ok := k.(common.Uint160); ok {
 			if list, ok := v.(*NonceSortedTxs); ok {
-				count := len(list.txs) + len(list.orphans)
-				programHashes[programHash] = count
+				if len := list.Len(); len != 0 {
```
yilunzhang: we should avoid using reserved word "len"

chain/pool/txpool.go
yilunzhang: We should also make sure that blockValidationState is not modified if processTx returns error

trueinsider: we can divide verify bvs into verify and append

trueinsider: or simply do clean in case of error

yilunzhang: Let's do this in a separate PR

PR #544, opened 2019-06-18
fix validation when adding tx to pool

trueinsider:
1. don't change block validation state until we're sure there is no errors
2. apply validation to nano pays too
3. add tx to tx maps when replacing old tx (remove old tx from maps too)
4. clean old tx from block validation state when replacing

### Checklist

### Extra information

### DIFF
--- a/chain/blockValidator.go
+++ b/chain/blockValidator.go
@@ -123,8 +123,10 @@ func TransactionCheck(ctx context.Context, block *block.Block) error {
 			return fmt.Errorf("transaction sanity check failed: %v", err)
 		}
 		if err := bvs.VerifyTransactionWithBlock(txn, block.Header); err != nil {
+			bvs.Reset()
 			return fmt.Errorf("transaction block check failed: %v", err)
 		}
+		bvs.Commit()
 		if err := VerifyTransactionWithLedger(txn, block.Header); err != nil {
 			return fmt.Errorf("transaction history check failed: %v", err)
 		}
--- a/chain/pool/txpool.go
+++ b/chain/pool/txpool.go
@@ -108,50 +108,65 @@ func (tp *TxnPool) processTx(txn *transaction.Transaction) error {
 		// sigchain txn should not be added to txn pool
 		return nil
 	case pb.NanoPayType:
+		tp.blockValidationState.Lock()
+		defer tp.blockValidationState.Unlock()
+		if err := tp.blockValidationState.VerifyTransactionWithBlock(txn, nil); err != nil {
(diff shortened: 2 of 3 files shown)


chain/blockValidator.go
```diff
@@ -123,8 +123,10 @@ func TransactionCheck(ctx context.Context, block *block.Block) error {
 			return fmt.Errorf("transaction sanity check failed: %v", err)
 		}
 		if err := bvs.VerifyTransactionWithBlock(txn, block.Header); err != nil {
+			bvs.Reset()
```
yilunzhang: Do we need this Reset at all? bvs is newly created and is never used

chain/pool/txpool.go
```diff
@@ -108,50 +108,65 @@ func (tp *TxnPool) processTx(txn *transaction.Transaction) error {
...
+
 			if err := list.Add(txn); err != nil {
 				return err
 			}
 
-			return nil
-		}
-
-		if list.Full() {
+			tp.deleteTransactionFromMap(txn)
```
yilunzhang: should this be oldTxn?

chain/pool/txpool.go
```diff
@@ -330,14 +345,26 @@ func (tp *TxnPool) CleanSubmittedTransactions(txns []*transaction.Transaction) e
...
 	}
 
-	if err := tp.blockValidationState.CleanSubmittedTransactions(txnsInPool); err != nil {
-		if err := tp.blockValidationState.RefreshBlockValidationState(tp.GetAllTransactions()); err != nil {
-			return err
-		}
+	return tp.CleanBlockValidationState(txnsInPool)
+}
+
+func (tp *TxnPool) addTransactionToMap(txn *transaction.Transaction) {
```
yilunzhang: Do you think we should put tp.NanoPayTxs.Store here?

chain/blockValidator.go
trueinsider: mainly for consistency sake, doesn't hurt anyway 😂

chain/pool/txpool.go
trueinsider: you're right

chain/pool/txpool.go
trueinsider: this is used for all kinds of txs, so I don't think so

chain/pool/txpool.go
trueinsider: fixed

NoelBright:
Looks good to me.

PR #547, opened 2019-06-19
Enhance security check for id generation and validation in sigchain and more

yilunzhang:
* Rename commit txn type to sigchain txn type
* Verify sigchain id before adding to porserver
* Validate BeneficiaryAddr on startup
* Change random beacon to vrf output + proof: random beacon now consists of 32 bytes of vrf output (unique) and 96 bytes of vrf proof (non unique). This makes id generation more secure (expected attack time increased from O(N) to O(N^2)) at the cost of 96 bytes per block.


api/common/interfaces.go
```diff
@@ -389,10 +389,10 @@ func getVersion(s Serverer, params map[string]interface{}) map[string]interface{
 	return respPacking(SUCCESS, config.Version)
 }
 
-// commitPor send por transaction
+// sendSigChain send por transaction
 // params: {"sigchain":<sigchain>}
 // return: {"resultOrData":<result>|<error data>, "error":<errcode>}
-func commitPor(s Serverer, params map[string]interface{}) map[string]interface{} {
+func sendSigChain(s Serverer, params map[string]interface{}) map[string]interface{} {
```
NoelBright: Is this APi necessary in future? It just for test.

yilunzhang: I don't think so, it's never been used expect for testing

NoelBright:
looks good to me.

chain/blockValidator.go
```diff
@@ -308,11 +307,11 @@ func SignerCheck(header *block.Header) error {
 
 	rawPubKey, err := crypto.DecodePoint(publicKey)
 	if err != nil {
-		return err
+		return fmt.Errorf("decode public key error: %v", err)
 	}
 	err = crypto.Verify(*rawPubKey, signature.GetHashForSigning(header), header.Signature)
 	if err != nil {
-		return err
+		return fmt.Errorf("invalid header signature %x", header.Signature)
```
gdmmx: @yilunzhang 
Suggestion:  Don't override err info from crypto.Verify. Just extend err info.

yilunzhang: You are right, let me correct it in the current PR

util/config/config.go
```diff
@@ -45,16 +47,19 @@ const (
...
 	MaxCompatibleProtocolVersion = 75
 	DefaultTxPoolCap             = 32
+	ShortHashSize                = uint32(8)
 )
 
 var (
-	ShortHashSize uint32 = 8
-	ShortHashSalt []byte = util.RandomBytes(32)
+	ShortHashSalt    = util.RandomBytes(32)
+	GenesisTimestamp = time.Date(2018, time.January, 0, 0, 0, 0, 0, time.UTC).Unix()
```
gdmmx: @yilunzhang 
fmt.Println(time.Date(2018, time.January, 0, 0, 0, 0, 0, time.UTC))
`2017-12-31 00:00:00 +0000 UTC`
Is it your expectation?

How about `2018-05-28`(first listed by exchange) or `2018-03-05`(first commit forked from OnChain DNA)?

yilunzhang: I was thinking maybe we should change this right before we launch mainnet so the timestamp of block 0 and 1 are close

PR #552, opened 2019-06-20
Override seed list before validate config and add genesisblockproposer cli argument

yilunzhang:

### DIFF
--- a/util/config/config.go
+++ b/util/config/config.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	gonat "github.com/nknorg/go-nat"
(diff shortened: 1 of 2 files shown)


util/config/config.go
```diff
@@ -176,6 +179,14 @@ func Init() error {
 		Parameters.BeneficiaryAddr = BeneficiaryAddr
 	}
 
+	if len(SeedList) > 0 {
+		Parameters.SeedList = strings.Split(SeedList, ",")
+	}
+
+	if len(GenesisBlockProposer) > 0 {
```
gdmmx: @yilunzhang  GenesisBlockProposer check should not only >0, but also should equal len(pubKey.Bytes())

yilunzhang: This is checked in some later commit

PR #554, opened 2019-06-20
V0.9: delete vm 

NoelBright:
delete vm 
simplify the contract module and Sign in wallet 


NoelBright:
Yes, I has considering remove contract and signature from vm and move to other folders. And Rename contract with other name.

api/common/interfaces.go
```diff
@@ -570,7 +570,7 @@ func getAddressByName(s Serverer, params map[string]interface{}) map[string]inte
 		return respPacking(INTERNAL_ERROR, err.Error())
 	}
 
-	scriptHash, err := program.CreateRedeemHash(pubKey)
+	scriptHash, err := program.CreateProgramHash(pubKey)
```
yilunzhang: should we also rename this `scriptHash` to `programHash`?

NoelBright: ok,I will

PR #558, opened 2019-06-23
Enhance the security of random id generation

yilunzhang:
* Make random beacon vrf depends only on prev random beacon vrf: the first 32 bytes of random beacon (vrf output) is now deterministic (unique) given block proposer such that block proposer cannot affect next vrf output by trying to manipulate block hash, thus making the vrf output at X blocks later exponentially hard to predict (in terms of X), because in order to predict it, one has to know the private key of next X block proposers.
* As a compensation to the lack of entropy in random beacon, the last 96 bytes of random beacon (vrf proof) now has 64 bytes of random source.
* GetID from multiple seeds: this will be helpful if a few seed nodes are faulty or malicious.
* Changed GenerateIDBlockDelay to 8 to further reduce the probability that an attacker can manipulate his id

Ps: this is an incompatible update as it changes consensus parameter. Do not merge until we are ready.


nknd.go
```diff
@@ -396,22 +396,41 @@ func GetID(seeds []string, publickey []byte) ([]byte, error) {
 		return id, nil
 	}
 
-	log.Infof("get ID from localnode faild, no ID in ledger: %v", err)
+	if err != nil {
+		log.Errorf("get ID from local ledger error: %v", err)
+	} else {
+		log.Infof("get no ID from local ledger")
+	}
```
gdmmx: How about `log.Infof("get ID from localnode return %v with err %v", id, err)` directly?

yilunzhang: We shouldn't log as info if there is an error. The default log level will soon be error

nknd.go
```diff
@@ -396,22 +396,41 @@ func GetID(seeds []string, publickey []byte) ([]byte, error) {
...
 
-		log.Warningf("get ID from %s met error: %v", seed, err)
+	for idStr, count := range counter {
+		if count > n/2 {
+			if idStr == "" {
+				return nil, nil
+			}
+			return []byte(idStr), nil
+		}
```
gdmmx: Why GetID from multi-seed and select the majority response?
Node will JoinNet fail with a fake ID got from malicious seed. And will restart nknd to request next random seed from seedList.

yilunzhang: He will not restart nknd if at least one of his neighbor is malicious and do not reject connection. Seed node can return id arbitrarily so if seed is malicious, he can guarantee the node that is joining has at least one malicious neighbor and thus will not restart.

On the other hand, we may want to provide a long seed list. The previous logic will cause the node to query every seed when id is not generated yet.

consensus/state.go
```diff
@@ -110,7 +110,9 @@ func (consensus *Consensus) getAllNeighborsConsensusState() (*sync.Map, error) {
 // consensus height, or zero if no majority can be found
 func (consensus *Consensus) getNeighborsMajorityConsensusHeight() uint32 {
 	for i := 0; i < getConsensusStateRetries; i++ {
-		time.Sleep(getConsensusStateRetryDelay)
+		if i > 0 {
+			time.Sleep(getConsensusStateRetryDelay)
```
gdmmx: How about move the sleep to end of `for` statement? Achieved the aim which ignore sleep of first neighbor but with no need for `if i>0` condition.

yilunzhang: This will make `continue` to continue immediately without retry delay

gdmmx: How about defer time.Sleep?

yilunzhang: Defer only works when function returns, this is a for loop

PR #559, opened 2019-06-24
move name service and pub/sub to statedb

trueinsider:
Move name service and pubsub storage into trie

### Checklist

### Extra information


chain/db/name.go
```diff
@@ -7,96 +7,153 @@ import (
...
-	// DELETE VALUE
-	err = cs.st.BatchDelete(registrantKey)
+func (sdb *StateDB) deleteName(registrantId string, nameId string) error {
+	err := sdb.trie.TryDelete(append(NameRegistrantPrefix, registrantId...))
 	if err != nil {
 		return err
 	}
-	err = cs.st.BatchDelete(nameKey)
+
+	err = sdb.trie.TryDelete(append(NamePrefix, nameId...))
```
yilunzhang: What if the previous TryDelete succeeded but this one failed?

trueinsider: we'll get incorrect state that won't be saved I guess. not sure what can be done about that.

chain/db/name.go
```diff
@@ -7,96 +7,153 @@ import (
...
+	for registrantId, name := range sdb.names {
+		nameId := getNameId(name)
+		registrant := sdb.nameRegistrants[nameId]
+		if name == "" {
+			sdb.deleteName(registrantId, nameId)
+		} else {
+			sdb.updateName(registrant, name)
+		}
+		if commit {
+			delete(sdb.names, registrantId)
```
yilunzhang: Why we need to remove these when `name != ""` ?

trueinsider: same logic as with account and nanopay. basically if we commit then we clean up.

yilunzhang: But I don't see FinalizeNames being called anywhere?

trueinsider: it's called in statedb.go/Finalize

chain/db/name.go
yilunzhang: I think it's better to provide some batch operation API in trie later

chain/db/pubsub.go
```diff
@@ -2,95 +2,220 @@ package db
...
-func (cs *ChainStore) Unsubscribe(subscriber []byte, identifier string, topic string, bucket uint32, duration uint32, height uint32) error {
-	if duration == 0 {
-		return nil
+func (ps *pubSub) Empty() bool {
+	return len(ps.subscriber) == 0 && len(ps.meta) == 0 && ps.expiresAt == 0
+}
+
+func getTopicId(topic string) []byte {
+	topicHash := sha256.Sum256([]byte(strings.ToLower(topic)))
+	return topicHash[:20]
```
yilunzhang: what about defining a constant instead of using hard coded 20 here?

trueinsider: 👍

chain/db/pubsub.go
```diff
@@ -2,95 +2,220 @@ package db
...
+
+	subscriberHash := sha256.Sum256(append(subscriber, identifier...))
+	pubSubId.Write(subscriberHash[:20])
+
+	return string(pubSubId.Bytes())
+}
+
+func getPubSubCleanupId(height uint32) string {
+	buf := new(bytes.Buffer)
+	_ = serialization.WriteUint32(buf, height)
```
yilunzhang: I think it's better to handle errors in getTopicBucketId, getPubSubId, getPubSubCleanupId (serialization.writeXXX and write) because in theory the error could happen, very unlikely but still possible

chain/db/pubsub.go
```diff
@@ -2,95 +2,220 @@ package db
...
+}
 
-	// DELETE VALUE
-	err := cs.st.BatchDelete(subscriberKey)
+func (sdb *StateDB) getPubSubCleanup(height uint32) (map[string]struct{}, error) {
+	var psc map[string]struct{}
+	var ok bool
+	if psc, ok = sdb.pubSubCleanup[height]; !ok {
+		enc, err := sdb.trie.TryGet(append(PubSubCleanupPrefix, getPubSubCleanupId(height)...))
+		if err != nil || len(enc) == 0 {
```
yilunzhang: I see that there is a `len(enc) == 0` here but not in `getPubSub`. Is there a reason for the difference?

chain/db/pubsub.go
trueinsider: I don't think error is possible if you look at buffer.go/line 177. it's nil everytime. error could happen depending on Writer, but we user Buffer here, so...

chain/db/pubsub.go
trueinsider: I think it's a bug (same on in nano pay), fixed

chain/db/pubsub.go
```diff
@@ -2,95 +2,220 @@ package db
...
+	id := getPubSubId(topic, bucket, subscriber, identifier)
 
-	err = cs.CancelKeyExpirationAtBlock(height+duration, subscriberKey)
+	ps, err := sdb.getPubSub(id)
 	if err != nil {
 		return err
 	}
 
+	if ps.Empty() {
+		ps.subscriber = address.MakeAddressString(subscriber, identifier)
```
yilunzhang: Do we need to store the full subscriber addr string here? Looks like identifier is enough?

trueinsider: why? it's the only place where subscriber is stored, there is no other place where I can get pk from

yilunzhang: Oh you are right. What about storing identifier (as string) and pk (as []byte) separately to save 32 bytes per subscriber?

chain/db/pubsub.go
```diff
@@ -99,27 +224,35 @@ func (cs *ChainStore) GetSubscribers(topic string, bucket uint32) map[string]str
 		identifier, _ := serialization.ReadVarString(rk)
 		subscriberString := address.MakeAddressString(subscriber, identifier)
 
-		subscribers[subscriberString] = string(iter.Value())
+		subscribers[subscriberString] = string(iter.Value)
```
yilunzhang: subscriberString should be stored in `iter.Value` right? The `iter.Key` now contains only the hash

chain/db/pubsub.go
trueinsider: done

chain/db/pubsub.go
trueinsider: yeah, noticed it too and fixed already :)

NoelBright:
there are some compile errors.

chain/db/pubsub.go
```diff
@@ -128,94 +261,120 @@ func (cs *ChainStore) GetFirstAvailableTopicBucket(topic string) int {
...
-		_, _ = serialization.ReadBytes(rk, uint64(len(prefix)))
+	rk := bytes.NewReader(lastKey)
 
-		bucket, _ := serialization.ReadUint32(rk)
+	// read prefix
+	_, _ = serialization.ReadBytes(rk, uint64(len(prefix)))
 
-		lastBucket = bucket
-	}
+	bucket, _ := serialization.ReadUint32(rk)
```
yilunzhang: 1. Instead of putting logic about key here, what about creating a function that parse bucket from key? Same for other places 
2. We may be able to ignore write error as you mentioned, but read error should definitely be handled because the data may be corrupted or in different format
3. Can we guarantee that the last key is the key that has the largest bucket? I think it's safer to parse every key and save the largest bucket id

trueinsider: 1. There is no other places though? ALso
3. It should be ordered, having to parse every key can be costly performance-wise.

NoelBright:
use "chain.DefaultLedger.Store.GetRegistrant(name)" to get public key, but got nil.

chain/db/pubsub.go
yilunzhang: Whether it's ordered depends on the serialization of Uint32. Specifically we use little endian, so it's not ordered. And even if we do, making this assumption here couples with the serialization encoding and is quite dangerous. On the other hand, parse this key should be very fast, and we only need to parse 1000 times at most

trueinsider:
@NoelBright how to reproduce?

NoelBright:
'./nknc name -r -name abcdefghmb -p tt --fee 0 --nonce 0' to register name.
and
curl -X POST  http://127.0.0.1:30003 -d '{ "method": "getaddressbyname", "params": { "name": "abcdefghmb" }, "id": 1 }'

to get address by name.

trueinsider:
@NoelBright have you registered the name before?

NoelBright:
> @NoelBright have you registered the name before?

Yes.

chain/db/pubsub.go
trueinsider: We need to parse 1000*1000 times at most, because it goes over every subscriber of every bucket.

trueinsider: 3. done

NoelBright:
there is same problem with pubsub. 

'./nknc subscribe -s --id abcdfeg --topic efg --bucket 1 --duration 10 --meta xyz --wallet wallet.json -p tt --fee 0 --nonce 1'

trueinsider:
@NoelBright try again please

NoelBright:
I found the codes below in StateDB sturcture are not be used:
	nanoPayCleanup  map[uint32]map[string]struct{}
	names           map[string]string
	nameRegistrants map[string][]byte
	pubSub          map[string]*pubSub
	pubSubCleanup   map[uint32]map[string]struct{}

trueinsider:
@NoelBright what do you mean?

NoelBright:
in "nknorg/nkn/chain/db/statedb.go".StateDB structure some members are not used,such as 
"
	nanoPay         map[string]*nanoPay
	nanoPayCleanup  map[uint32]map[string]struct{}
	names           map[string]string
	nameRegistrants map[string][]byte
	pubSub          map[string]*pubSub
	pubSubCleanup   map[uint32]map[string]struct{}
"

trueinsider:
@NoelBright but they're used though?

yilunzhang:
Looks good to me

api/common/interfaces.go
```diff
@@ -601,7 +604,10 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
 		return respPacking(INVALID_PARAMS, "bucket should be a string")
 	}
 
-	subscribers := chain.DefaultLedger.Store.GetSubscribers(topic, uint32(bucket))
+	subscribers, err := chain.DefaultLedger.Store.GetSubscribers(topic, uint32(bucket))
```
NoelBright: should deal with the case : subscribers  is a null map  'map[]'

NoelBright:
Looks good to me.

api/common/interfaces.go
yilunzhang: Now client side will get an empty map if subscribers is nil, which is expected IMO

PR #563, opened 2019-06-26
V0.9: add issueAsset structure and modify account structure 

NoelBright:
add issueAsset structure and modify account structure 


chain/db/account.go
```diff
@@ -80,16 +173,20 @@ func (acc *account) SetNonce(nonce uint64) {
 	acc.nonce = nonce
 }
 
-func (acc *account) SetBalance(balance common.Fixed64) {
-	acc.balance = balance
+func (acc *account) SetBalance(assetID common.Uint256, symbol string, amount common.Fixed64) {
+	acc.balances[assetID] = &balance{
```
yilunzhang: I was thinking that, if `acc.balances[assetID]` already exists, should we just setting the value of each field in `acc.balances[assetID]` to avoid creating a new struct each time balance is changed?

chain/db/issue.go
```diff
@@ -0,0 +1,149 @@
...
+type issueAsset struct {
+	name        string
+	symbol      string
+	totalSupply common.Fixed64
+	precision   byte
+	owner       common.Uint160
+}
+
+func (ia *issueAsset) Empty() bool {
+	return len(ia.name) == 0 && len(ia.symbol) == 0
```
yilunzhang: Should we also check other fields in `Empty()`?

chain/db/issue.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+
+	sdb.assets[assetID] = asset
+
+	return asset, nil
+}
+
+func (sdb *StateDB) deleteAsset(assetID common.Uint256) error {
+	delete(sdb.assets, assetID)
```
yilunzhang: we didn't remove asset from trie here

NoelBright: asset need not be delete, it's can only be set & read.

NoelBright:
> I have 2 questions:
> 
> 1. Do we need to use uint256 for assetid? Seems that we don't need to consider hash collision and uint64 or even uint32 should be enough
> 2. Asset symbol seems to be not used, while in updateBalance it's able to modify it
> 3. It seems possible to have multiple assets with the same symbol

1. assetid will be set with the issue token txid.
2. asset symbol could be used to get asset by symbol. we can append symbol with a checksum of assetid to avoid the same symbol.

chain/db/issue.go
NoelBright: 0 total supply and  0 precision is legal.

chain/db/account.go
NoelBright: I will update this function.

chain/db/issue.go
yilunzhang: But is it valid that a asset has empty name and symbol?

yilunzhang:
> 2. asset symbol could be used to get asset by symbol. we can append symbol with a checksum of assetid to avoid the same symbol.

It seems that asset symbol is mutable? Otherwise we shouldn't pass symbol as argument in SetBalance()

chain/db/issue.go
NoelBright: no, asset must have name and symbol.

NoelBright:
> > 1. asset symbol could be used to get asset by symbol. we can append symbol with a checksum of assetid to avoid the same symbol.
> 
> It seems that asset symbol is mutable? Otherwise we shouldn't pass symbol as argument in SetBalance()

sure. should check it.

PR #564, opened 2019-06-27
V0.9 dashboard node status

iheron:

### Checklist

### Extra information


Makefile
```diff
@@ -12,18 +12,27 @@ IDENTIFIER=$(GOOS)-$(GOARCH)
 help:  ## Show available options with this Makefile
 	@grep -F -h "##" $(MAKEFILE_LIST) | grep -v grep | awk 'BEGIN { FS = ":.*?##" }; { printf "%-15s  %s\n", $$1,$$2 }'
 
+web:
+	@cd dashboard/web && yarn build && cp -a ./dist ../../web
+
+yarn:
+	@rm -rf web
+	@cd dashboard/web && yarn build && cp -a ./dist ../../web
```
yilunzhang: If this is intended to be the same as `web`, we can use `make web` for this line

Makefile
```diff
@@ -12,18 +12,27 @@ IDENTIFIER=$(GOOS)-$(GOARCH)
 help:  ## Show available options with this Makefile
 	@grep -F -h "##" $(MAKEFILE_LIST) | grep -v grep | awk 'BEGIN { FS = ":.*?##" }; { printf "%-15s  %s\n", $$1,$$2 }'
 
+web:
+	@cd dashboard/web && yarn build && cp -a ./dist ../../web
+
+yarn:
```
yilunzhang: We need to add `.PHONY: yarn` to tell Makefile that it will not try to find the file/directory named `yarn`. See `build` below for example

Makefile
yilunzhang: What's the difference between `make web` and `make yarn` by design?

Makefile
```diff
@@ -12,18 +12,27 @@ IDENTIFIER=$(GOOS)-$(GOARCH)
...
 .PHONY: crossbuild
-crossbuild:
+crossbuild: web
 	mkdir -p build/$(IDENTIFIER)
 	make build FLAGS="build/$(IDENTIFIER)"
+	@cp -a dashboard/web/dist build/$(IDENTIFIER)/web
 
 .PHONY: all
 all:  ## Build binaries for all available architectures
+	@rm -rf web
```
yilunzhang: We do we rm web here? If we just want a fresh build, I think we can use `make yarn` right?

Makefile
```diff
@@ -12,18 +12,27 @@ IDENTIFIER=$(GOOS)-$(GOARCH)
...
+web:
+	@cd dashboard/web && yarn build && cp -a ./dist ../../web
+
+yarn:
+	@rm -rf web
+	@cd dashboard/web && yarn build && cp -a ./dist ../../web
+
 .PHONY: build
-build:
+build:
```
yilunzhang: We should also add `web` as build dependencies because Docker will use `make build` directly

Makefile
```diff
@@ -36,7 +45,7 @@ all:  ## Build binaries for all available architectures
 	make crossbuild GOOS=windows GOARCH=386
 
 .PHONY: build_local
-build_local:  ## Build local binaries without providing specific GOOS/GOARCH
+build_local: ## Build local binaries without providing specific GOOS/GOARCH
```
yilunzhang: We should also add `web` as build_local dependencies because makefile will use `make build_local` by default

dashboard/routes/wallet/status.go
```diff
@@ -1,23 +1,80 @@
...
 		if exists {
-			context.JSON(http.StatusOK, wallet)
+			account, err := wallet.(vault.Wallet).GetDefaultAccount()
+			if err != nil {
+				log.WebLog.Error("get wallet account error: ", err)
+				context.AbortWithError(http.StatusInternalServerError, err)
+				return
+			}
+
+			address, err := account.ProgramHash.ToAddress()
```
yilunzhang: We can use `account.GetBalance()` directly

dashboard/routes/wallet/status.go
```diff
@@ -1,23 +1,80 @@
...
+		if exists {
+			account, err := wallet.(vault.Wallet).GetDefaultAccount()
+			if err != nil {
+				log.WebLog.Error("get wallet account error: ", err)
+				context.AbortWithError(http.StatusInternalServerError, err)
+				return
+			}
+
+			context.JSON(http.StatusOK, gin.H{
+				"privateKey": BytesToHexString(account.PrivateKey),
```
yilunzhang: We should use `crypto.GetSeedFromPrivateKey(account.PrivateKey)` instead of `account.PrivateKey`.
@NoelBright We need also to change it in `showAccountInfo` because seed is what's used to recover a wallet in sdk

Makefile
iheron: make all: first `@rm -rf web` then build `web` once
make yarn: `@rm -rf web` only build `web`, not build `nknc` `nknd`
if `@rm -rf web` then `make all` will build `web` before every platform.
I don't know how to code `make web`. Could you help me?

Makefile
gdmmx: yarn is a PHONY target.
We don't expect to `yarn build` repeatedly for each crossPlatform during `make all`

gdmmx: On the other hand, `yarn build` just only overwrite the output files with same path. It would not delete those files which was obsolete.

Makefile
gdmmx: We need two make target:
* rebuild web dir with no excuse
* rebuild web dir only case in resource files was updated

iheron:
Just merge #567 , close this

PR #565, opened 2019-06-27
add issueAsset structure and modify account structure

NoelBright:
add issueAsset structure and modify account structure

    1. add issueAsset structure
    2. modify account structure
    3. add assetid for NKN asset.


pb/transaction.proto
```diff
@@ -99,3 +100,11 @@ message NanoPay {
 	uint32 height   = 5;
 	uint32 duration = 6;
 }
+
+message IssueAsset {
+	bytes sender       = 1;
+	string name        = 2;
+	string symbol      = 3;
+    int64 total_supply = 4;
```
yilunzhang: indent seems to be wrong

chain/db/issue.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := asset.Deserialize(buff); err != nil {
+		return nil, err
+	}
+
+	sdb.assets[assetID] = asset
+
+	return asset, nil
+}
+
+func (sdb *StateDB) deleteAsset(assetID common.Uint256) error {
```
yilunzhang: It seems that we don't need `deleteAsset` function at all?

chain/db/issue.go
```diff
@@ -0,0 +1,182 @@
...
+			sdb.updateAsset(assetID, asset)
+		}
+
+		if commit {
+			delete(sdb.assets, assetID)
+		}
+	}
+}
+
+func (cs *ChainStore) GetAsset(assetID common.Uint256) (name, symbol string, totalSupply common.Fixed64, precision uint32, err error) {
```
yilunzhang: What about making issueAsset a public type? Also `Asset` seems a better name

chain/txValidator.go
```diff
@@ -197,6 +198,36 @@ func CheckTransactionPayload(txn *transaction.Transaction) error {
...
 			return errors.New(fmt.Sprintf("duration can't be lower than %d", transaction.MinNanoPayDuration))
 		}
+
+	case pb.ISSUE_ASSET_TYPE:
+		pld := payload.(*pb.IssueAsset)
+		if len(pld.Sender) != UINT160SIZE {
+			return errors.New("length of programhash error")
+		}
+
+		match, err := regexp.MatchString("(^[a-z][A-Za-z0-9 ]{2,11}$)", pld.Name)
```
yilunzhang: We also need to change the first char (to allow A-Z)

chain/txValidator.go
```diff
@@ -197,6 +198,36 @@ func CheckTransactionPayload(txn *transaction.Transaction) error {
...
+		if err != nil {
+			return err
+		}
+		if !match {
+			return fmt.Errorf("name %s should start with a letter, contain a-z0-9 and have length 3-12", pld.Name)
```
yilunzhang: Also description needs to be updated

chain/db/issue.go
NoelBright: Asset is better than issueAsset, but this function is not suitable to return Asset. Because the db module is referenced by interface LedgerStore.

chain/db/issue.go
NoelBright: it's can be used to delete invalid asset in statedb.

PR #569, opened 2019-06-28
Add genesis signer id into genesis block and rename Signer/ChordId to SignerPk/SignerId in block header

yilunzhang:
* Rename Signer/ChordId to SignerPk/SignerId in block header
* Add genesis signer id into genesis block
* Add license for trie module 


NoelBright:
looks good to me.

PR #573, opened 2019-06-30
verify tx with ledger when proposing a block

trueinsider:
There is times when tx (already invalid) getting added to a proposed block. Block proposer should verify such txs and don't add them to the block.

### Checklist

### Extra information

### DIFF
--- a/chain/mining.go
+++ b/chain/mining.go
@@ -65,6 +65,8 @@ func (bm *BuiltinMining) BuildBlock(ctx context.Context, height uint32, chordID 
 		return nil, err
 	}
 
+	bvs := NewBlockValidationState()
+
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


yilunzhang:
Since this is a compatible change, we will test and merge it after mainnet launch :)

NoelBright:
Looks good to me.

chain/txValidator.go
```diff
@@ -488,15 +487,13 @@ func (bvs *BlockValidationState) VerifyTransactionWithBlock(txn *transaction.Tra
...
-		if header != nil {
-			coinbase := payload.(*pb.Coinbase)
-			donationAmount, err := DefaultLedger.Store.GetDonation()
-			if err != nil {
-				return err
-			}
-			if Fixed64(coinbase.Amount) != GetRewardByHeight(header.UnsignedHeader.Height)+donationAmount {
-				return errors.New("Mining reward incorrectly.")
-			}
+		coinbase := payload.(*pb.Coinbase)
```
yilunzhang: Why not skipping the check when height is 0?

chain/mining.go
```diff
@@ -93,17 +95,30 @@ func (bm *BuiltinMining) BuildBlock(ctx context.Context, height uint32, chordID
...
+			continue
+		}
+		if err := VerifyTransactionWithLedger(txn); err != nil {
+			log.Warning("invalid transaction")
+			txnCollection.Pop()
+			continue
+		}
+		if err := bvs.VerifyTransactionWithBlock(txn, height); err != nil {
+			bvs.Reset()
+			log.Warning("invalid transaction")
```
yilunzhang: It's better to use `log.Warningf("invalid transaction: %v", err)` here and in L104

chain/txValidator.go
trueinsider: height is only 0 when called from txpool and txpool don't have coinbase txs

yilunzhang: I think we should further make sure that by adding a switch condition in pool.processTx that returns error when type is coinbase

PR #575, opened 2019-07-02
Disable register name txn temporarily and add verification when proposing block

yilunzhang:
* Disable register name txn temporarily due to the recent abuse
* Verify txn verification when proposing block and remove invalid txn from pool


gdmmx:
timestamp commit 34c803a OK👌

PR #593, opened 2019-07-18
Increase robustness to extremely slow nodes

yilunzhang:
Upgrade to the latest nnet version to achieve:
* Increase measured RTT if ping msg timeout: This will avoid packets being routed to recently timeout neighbor
* Update lastRxTime only when receiving ping message: This may fix the case when remote node has extremely slow upstream bandwidth and all message timeout, but it still stays in other nodes' neighbors.
* ToRemoteNodeList only returns remote nodes that are not stopped


yilunzhang:
> I think the piece of `update rn.lastRxTime` in `nnet/node/remotenode.go` should move to ahead of rn.handleMsgBuf.

Reduced ping retry interval when timeout as we discussed

PR #594, opened 2019-07-19
Web ui

iheron:

### Checklist

### Extra information


util/config/config.go
```diff
@@ -431,3 +436,33 @@ func (config *Configuration) CheckPorts(myIP string) (bool, error) {
...
+func OpenConfigFile() ([]byte, error) {
+	var file []byte
+
+	configFile := ConfigFile
+	if configFile == "" {
+		configFile = DefaultConfigFile
+	}
+	_, err := os.Stat(configFile)
+	if err != nil {
+		return file, err
```
yilunzhang: We should return `nil, error` here to make sure we don't accidentally return non-nil file if error is not nil

util/config/config.go
```diff
@@ -431,3 +436,33 @@ func (config *Configuration) CheckPorts(myIP string) (bool, error) {
...
+		return nil, err
+	}
+
+	// Remove the UTF-8 Byte Order Mark
+	file = bytes.TrimPrefix(file, []byte("\xef\xbb\xbf"))
+	return file, err
+
+}
+
+func WriteConfigFile(data []byte) error {
```
yilunzhang: It's better to accept a config as input argument, and put `bytes, err := json.MarshalIndent(&configuration, "", "    ")` part inside this function

util/config/config.go
```diff
@@ -431,3 +436,33 @@ func (config *Configuration) CheckPorts(myIP string) (bool, error) {
 	}
 	return true, nil
 }
+
+func OpenConfigFile() ([]byte, error) {
+	var file []byte
+
+	configFile := ConfigFile
```
yilunzhang: This get config file name part duplicates in OpenConfigFile() and WriteConfigFile(), we can write a common function for this part, such that later when we need to change it, we only need to change it once

dashboard/routes/wallet/create.go
```diff
@@ -33,17 +36,73 @@ func WalletCreateRouter(router *gin.RouterGroup) {
 			log.WebLog.Error("wallet file exists.")
 			context.AbortWithError(http.StatusInternalServerError, errors.New("wallet file exists."))
 			return
-		} else {
-			_, err := vault.NewWallet(config.Parameters.WalletFile, []byte(data.Password), true)
+		}
+
+		if config.Parameters.WebGuiCreateWallet && !config.Parameters.AllowEmptyBeneficiaryAddress {
+			if data.BeneficiaryAddr == "" {
```
yilunzhang: I think we should move `!config.Parameters.AllowEmptyBeneficiaryAddress` from the previous line to this line, so that even if `config.Parameters.AllowEmptyBeneficiaryAddress == true`, we can still set beneficiary address

dashboard/routes/wallet/create.go
```diff
@@ -33,17 +36,73 @@ func WalletCreateRouter(router *gin.RouterGroup) {
...
+				log.WebLog.Error("beneficiary address is empty.")
+				context.AbortWithError(http.StatusBadRequest, errors.New("beneficiary address is empty."))
+				return
+			}
+
+			_, err = ToScriptHash(data.BeneficiaryAddr)
```
yilunzhang: From L48 to L85 is the same as set beneficiary in BeneficiaryRouter, what about making a common function so both place can simply call it?

dashboard/routes/wallet/open.go
```diff
@@ -40,6 +40,13 @@ func WalletOpenRouter(router *gin.RouterGroup) {
 				context.AbortWithError(http.StatusForbidden, err)
 				return
 			}
+
+			err = password.SavePassword()
```
yilunzhang: Why do we need to save password when opening wallet here?

nknd.go
```diff
@@ -156,16 +156,30 @@ func nknMain(c *cli.Context) error {
 
 	var wallet vault.Wallet
 	var account *vault.Account
-	for wallet == nil {
-		time.Sleep(time.Second * 3)
+	if config.Parameters.WebGuiCreateWallet {
+		for wallet == nil {
```
yilunzhang: I think it's better to use `wallet == nil || account == nil` just in case get wallet success but get account fails

util/password/password.go
```diff
@@ -70,3 +83,37 @@ func GetAccountPassword() ([]byte, error) {
...
 }
+
+func ReadPasswordFile() ([]byte, error) {
+	var file []byte
+
+	passwordFile := config.Parameters.PasswordFile
+
+	_, err := os.Stat(passwordFile)
+	if err != nil {
+		return file, err
```
yilunzhang: It's always good to explicitly return `nil, err` when `err != nil`

util/config/config.go
```diff
@@ -431,3 +436,33 @@ func (config *Configuration) CheckPorts(myIP string) (bool, error) {
...
+		return file, err
+	}
+	file, err = ioutil.ReadFile(configFile)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the UTF-8 Byte Order Mark
+	file = bytes.TrimPrefix(file, []byte("\xef\xbb\xbf"))
+	return file, err
```
yilunzhang: It's better to explicitly return `file, nil` here so people know error is nil

util/password/password.go
```diff
@@ -70,3 +83,37 @@ func GetAccountPassword() ([]byte, error) {
...
+	}
+
+	file, err = ioutil.ReadFile(passwordFile)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the UTF-8 Byte Order Mark
+	file = bytes.TrimPrefix(file, []byte("\xef\xbb\xbf"))
+	return file, err
```
yilunzhang: It's better to explicitly return `file, nil` here so people know error is nil

util/password/password.go
```diff
@@ -70,3 +83,37 @@ func GetAccountPassword() ([]byte, error) {
...
+}
+
+func SavePassword() error {
```
yilunzhang: It's better to pass password to this function because Passwd is not always guaranteed to contain password (unless passed in through -p argument or created in dashboard)

dashboard/routes/wallet/create.go
```diff
@@ -92,12 +92,7 @@ func WalletCreateRouter(router *gin.RouterGroup) {
 			return
 		}
 		password.Passwd = data.Password
-		err = password.SavePassword()
-		if err != nil {
-			log.WebLog.Error("save wallet password error: ", err)
-			context.AbortWithError(http.StatusInternalServerError, err)
-			return
-		}
+		password.SavePassword()
```
yilunzhang: Why not handling error here?

dashboard/routes/wallet/open.go
```diff
@@ -27,29 +42,34 @@ func WalletOpenRouter(router *gin.RouterGroup) {
...
-				log.WebLog.Error("save wallet error: ", err)
-				context.AbortWithError(http.StatusInternalServerError, err)
+				log.WebLog.Error("open wallet error: ", err)
+				context.AbortWithError(http.StatusForbidden, err)
 				return
 			}
-			context.JSON(http.StatusOK, "")
-			return
+
+			password.SavePassword()
```
yilunzhang: Same as above, why not handling error here?

dashboard/web/store/wallet.js
```diff
@@ -39,6 +40,7 @@ const actions = {
   async createWallet({commit, rootState}, {password, beneficiaryAddr}) {
     try {
       let res = await this.$axios.post('/api/wallet/create', {password: password, beneficiaryAddr:beneficiaryAddr})
+      localStorage.setItem('seed', seedHash(password))
```
yilunzhang: Are we storing password in local storage? This seems to be very dangerous as anyone who has access to the computer will now be able to see the password plaintext.

dashboard/routes/wallet/open.go
iheron: if wallet file exists, --password-file parameters can work?

yilunzhang: I think it's better to save password to file only when creating wallet but not opening wallet, otherwise if user used dashboard once, the password plaintext will be saved to disk...

util/config/config.go
```diff
@@ -459,14 +464,50 @@ func OpenConfigFile() ([]byte, error) {
...
+	if err != nil {
+		return err
 	}
-	return ioutil.WriteFile(configFile, data, 0666)
+	configFile := GetConfigFile()
+	return ioutil.WriteFile(configFile, bytes, 0666)
+}
+
+func SetBeneficiaryAddr(addr string, allowEmpty bool) error {
+	if allowEmpty {
```
yilunzhang: Seems that this should be `!allowEmpty`

util/config/config.go
```diff
@@ -459,14 +464,50 @@ func OpenConfigFile() ([]byte, error) {
...
+	if addr != "" {
+		_, err := ToScriptHash(addr)
+		if err != nil {
+			return errors.New(fmt.Sprintf("parse BeneficiaryAddr error: %v", err))
+		}
+	}
+
+	file, err := OpenConfigFile()
+	if err != nil {
+		return errors.New("Config file not exists.")
```
yilunzhang: It's better also to return the err here, e.g. `fmt.Errorf("open config file error: %v", err)` or simply just return err

PR #603, opened 2019-08-01
Limit Subscribe txn string length

gdmmx:

Urgent bugfix temporary: limit Subscribe txn string length

### Checklist

### Extra information

### DIFF
--- a/chain/txValidator.go
+++ b/chain/txValidator.go
@@ -165,6 +165,15 @@ func CheckTransactionPayload(txn *transaction.Transaction) error {
 		if !match {
 			return fmt.Errorf("topic %s should start with a letter, contain A-Za-z0-9-_.+ and have length 3-255", topic)
 		}
+
+		// Check sub.Meta & sub.Identifier limitation via height
+		h := DefaultLedger.Store.GetHeight() + 1 // txn's height should be currHeight + 1
+		if len(pld.Identifier) > config.MaxTxnSubIdentifierList.GetValueAtHeight(h) {
+			return errors.New("Identifier too long")
+		}
+		if len(pld.Meta) > config.MaxTxnSubMetaList.GetValueAtHeight(h) {
+			return errors.New("Meta too long")
+		}
 	case pb.GENERATE_ID_TYPE:
 		pld := payload.(*pb.GenerateID)
 		_, err := crypto.NewPubKeyFromBytes(pld.PublicKey)
--- a/nknd.go
+++ b/nknd.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
-	NetVersionNum = 3 // This is temporary and will be removed soon after mainnet is stabilized
+	NetVersionNum = 4 // This is temporary and will be removed soon after mainnet is stabilized
 )
 
 var (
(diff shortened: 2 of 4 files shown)


pb/transaction.go
```diff
@@ -135,3 +136,24 @@ func (m *SigChainTxn) ToMap() map[string]interface{} {
 		"Submitter": common.BytesToUint160(m.Submitter),
 	}
 }
+
+type SubTxnDBChangePoint uint32
+
+const (
+	LimitMetaCP SubTxnDBChangePoint = 124000
```
NoelBright: How about move this const SubTxnDBChangePoint to config.go?

yilunzhang:
What about this:
in config.go add 2 constants `maxAddrIdentifierSize` and `maxSubscribeMetaSize`, and 2 methods `GetMaxAddrIdentifierSize(height)` and `GetMaxSubscribeMetaSize(height)`, and return corresponding constant depending on height. This way we are separating verification logic (chain/txValidator.go) and constant definition (config.go).

gdmmx:
> What about this:
> in config.go we add 2 constants (for latest value) `maxAddrIdentifierSize = 64` and `maxSubscribeMetaSize = 1024`, and 2 methods `GetMaxAddrIdentifierSize(height)` and `GetMaxSubscribeMetaSize(height)`, and return corresponding constant depending on height. This way we are separating verification logic (chain/txValidator.go) and constant definition (config.go).

We have to consider that `ChangePoint`(likes `CheckPoint`) not only once in future, I think
`ChangePoint` should be a history Data, not a process routine. For readability and clear eachChange in history.

yilunzhang:
All different GetXXXX() should not be put into the same function, otherwise we will end up with tons of if...else... after a few updates, and it's very hard to see the history of each parameter.
To make the history data explicit and understandable, I think it's better to have something like:
```
GetA(height) {
if (height < xxx) {return xxx}
if (height < xxx) {return xxx}
return a
}

GetB(height) {
if (height < xxx) {return xxx}
if (height < xxx) {return xxx}
return b
}
```
Then we get a very clear and human understandable parameter history.

Conceptually these functions are just extending constants in config.go from static to dynamic, so it's better to put it in config also. Of course we can put it in another file in util/config but still in the same go pkg.

gdmmx:
Now we have two implements:

**Impl_1**
```
GetA(height) {
if (height < xxx) {return xxx}
if (height < xxx) {return xxx}
return a
}

GetB(height) {
if (height < xxx) {return xxx}
if (height < xxx) {return xxx}
return b
}
```
And 
**Impl_2**
```
A.Get(height) {
if (height < config.ChangePoint.A.xxx) {return xxx}
if (height < config.ChangePoint.A.xxx) {return xxx}
return a
}
B.Get(height) {
if (height < config.ChangePoint.B.xxx) {return xxx}
if (height < config.ChangePoint.B.xxx) {return xxx}
return a
}
```

Which is your preference?

yilunzhang:
I think this is probably a more generic way:
```
type HeightDependentInt struct {
  heights []uint32
  values []int
}

func (hdt *HeightDependentInt) ValueAtHeight(height uint32) int {
  return hdt.values[i] if height > hdt.heights[i]
}

A = HeightDependentInt{
  heights: []uint32{100000, 0},
  values: []int{100, 200},
}

A.ValueAtHeight(12345)
```

pb/transaction.go
gdmmx: Done.

NoelBright:
Looks good to me.

gdmmx:
Done as review suggestion.

PR #605, opened 2019-08-01
Print error and exit instead of panic directly

yilunzhang:
The helps prevent the release binary printing system absolute path in panic stacktrace. I've tested and found that existing options to remove it (e.g. GOROOT_FINAL) is not guaranteed to work in all cases and Golang does not provide a way to build bitwise identical binary in different environment yet. So the most robust way for now is to capture panic and print the panic message and then exit instead of panic directly. This should not hide stack trace info as panic msg are almost unique now.


NoelBright:
Looks good to me.

PR #609, opened 2019-08-04
add pubsub enchancements

trueinsider:
https://forum.nkn.org/t/nkp-0011-extend-subscription-and-unsubscribe/1483/6
1. ability to prolong existing subscription (doesn't error anymore)
2. `getsubscription` rpc method to get subscription duration and meta. use to decide if it's time to extend
3. unsubscribe

https://forum.nkn.org/t/nkp-0007-include-subscribers-in-txn-pool-in-getsubscribers-rpc-response/1441/2
1. get subscribers from pool too (doesn't return meta anymore, use `getsubscription` to get meta)

Deprecate buckets:
1. make bucket optional (0 by default)
2. increase bucket limit to 100k (because in the future we'll use 0 bucket all the time)
3. specify offset and limit when requesting subscribers (by default will return all subscribers in a bucket for compatibility reasons, but in the future you **MUST** specify `offset` and `limit`)
4. removed `getfirstavailabletopicbucket` and `gettopicbucketscount`

Bugfixes:
1. fix subscription count BVS (uncommited changes didn't count towards subscriptions limit)

### Type

### Checklist

### DIFF
--- a/api/common/interfaces.go
+++ b/api/common/interfaces.go
@@ -634,8 +634,52 @@ func getAddressByName(s Serverer, params map[string]interface{}) map[string]inte
 	return respPacking(SUCCESS, address)
 }
 
+// getSubscription get subscription
+// params: {"topic":<topic>, "bucket":<bucket>, "subscriber":<subscriber>}
+// return: {"resultOrData":<result>|<error data>, "error":<errcode>}
+func getSubscription(s Serverer, params map[string]interface{}) map[string]interface{} {
+	if len(params) < 2 {
+		return respPacking(INVALID_PARAMS, "length of params is less than 2")
+	}
+
+	topic, ok := params["topic"].(string)
+	if !ok {
+		return respPacking(INVALID_PARAMS, "topic should be a string")
+	}
+
+	var bucket float64
+	if _, ok := params["bucket"]; ok {
+		bucket, ok = params["bucket"].(float64)
+		if !ok {
+			return respPacking(INVALID_PARAMS, "bucket should be a float64")
+		}
+	}
+
+	subscriber, ok := params["subscriber"].(string)
+	if !ok {
+		return respPacking(INVALID_PARAMS, "subscriber should be a string")
+	}
+
+	_, pubKey, identifier, err := address.ParseClientAddress(subscriber)
+	if err != nil {
+		return respPacking(INTERNAL_ERROR, err.Error())
+	}
+
+	meta, expiresAt, err := chain.DefaultLedger.Store.GetSubscription(topic, uint32(bucket), pubKey, identifier)
+	if err != nil {
+		return respPacking(INTERNAL_ERROR, err.Error())
+	}
+	return respPacking(SUCCESS, struct {
+		Meta      string `json:"meta"`
+		ExpiresAt uint32 `json:"expiresAt"`
+	}{
+		meta,
+		expiresAt,
+	})
+}
+
 // getSubscribers get subscribers by topic
-// params: {"topic":<topic>, "bucket":<bucket>}
+// params: {"topic":<topic>, "bucket":<bucket>, "offset":<offset>, "limit":<limit>, "meta":<meta>, "txPool":<txPool>}
 // return: {"resultOrData":<result>|<error data>, "error":<errcode>}
 func getSubscribers(s Serverer, params map[string]interface{}) map[string]interface{} {
 	if len(params) < 2 {
@@ -647,16 +691,87 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
 		return respPacking(INVALID_PARAMS, "topic should be a string")
 	}
 
-	bucket, ok := params["bucket"].(float64)
-	if !ok {
-		return respPacking(INVALID_PARAMS, "bucket should be a string")
+	var bucket float64
+	if _, ok := params["bucket"]; ok {
+		bucket, ok = params["bucket"].(float64)
+		if !ok {
+			return respPacking(INVALID_PARAMS, "bucket should be a float64")
+		}
 	}
 
-	subscribers, err := chain.DefaultLedger.Store.GetSubscribers(topic, uint32(bucket))
+	var offset float64
+	if _, ok := params["offset"]; ok {
+		offset, ok = params["offset"].(float64)
+		if !ok {
+			return respPacking(INVALID_PARAMS, "offset should be a float64")
+		}
+	}
+
+	var limit float64
+	if _, ok := params["limit"]; ok {
+		limit, ok = params["limit"].(float64)
+		if !ok {
+			return respPacking(INVALID_PARAMS, "limit should be a float64")
+		}
+	} else {
+		limit = 1000
+	}
+
+	txPool, _ := params["txPool"].(bool)
+
+	localNode, err := s.GetNetNode()
 	if err != nil {
 		return respPacking(INTERNAL_ERROR, err.Error())
 	}
-	return respPacking(SUCCESS, subscribers)
+
+	response := make(map[string]interface{})
+
+	meta, _ := params["meta"].(bool)
+	var subscribers interface{}
+	if !meta {
+		subscribers, err = chain.DefaultLedger.Store.GetSubscribers(topic, uint32(bucket), uint32(offset), uint32(limit))
+	} else {
+		subscribers, err = chain.DefaultLedger.Store.GetSubscribersWithMeta(topic, uint32(bucket), uint32(offset), uint32(limit))
+	}
+	if err != nil {
+		return respPacking(INTERNAL_ERROR, err.Error())
+	}
+	response["subscribers"] = subscribers
+
+	if txPool {
+		if !meta {
+			response["subscribersInTxPool"] = localNode.GetTxnPool().GetSubscribers(topic)
+		} else {
+			response["subscribersInTxPool"] = localNode.GetTxnPool().GetSubscribersWithMeta(topic)
+		}
+	}
+
+	return respPacking(SUCCESS, response)
+}
+
+// getSubscribersCount get subscribers count by topic
+// params: {"topic":<topic>, "bucket":<bucket>}
+// return: {"resultOrData":<result>|<error data>, "error":<errcode>}
+func getSubscribersCount(s Serverer, params map[string]interface{}) map[string]interface{} {
+	if len(params) < 1 {
+		return respPacking(INVALID_PARAMS, "length of params is less than 1")
+	}
+
+	topic, ok := params["topic"].(string)
+	if !ok {
+		return respPacking(INVALID_PARAMS, "topic should be a string")
+	}
+
+	var bucket float64
+	if _, ok := params["bucket"]; ok {
+		bucket, ok = params["bucket"].(float64)
+		if !ok {
+			return respPacking(INVALID_PARAMS, "bucket should be a float64")
+		}
+	}
+
+	count := chain.DefaultLedger.Store.GetSubscribersCount(topic, uint32(bucket))
+	return respPacking(SUCCESS, count)
 }
 
 // getAsset get subscribers by topic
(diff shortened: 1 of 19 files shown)


trueinsider:
It says breaking changes, but actually the only thing that's breaking is `getsubscribers` rpc response and removed obsolete bucket-related RPCs

api/common/interfaces.go
```diff
@@ -647,91 +647,152 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
 		return respPacking(INVALID_PARAMS, "topic should be a string")
 	}
 
-	bucket, ok := params["bucket"].(float64)
+	var bucket float64
+	if _, ok := params["bucket"]; ok {
+		bucket, ok = params["bucket"].(float64)
```
yilunzhang: Is there a reason we are not checking uint32 here? Also do we keep bucket because of compatibility?

api/common/interfaces.go
```diff
@@ -647,91 +647,152 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
...
+			return respPacking(INVALID_PARAMS, "bucket should be a float64")
+		}
 	}
 
-	assetID, err := common.Uint256ParseFromBytes(hexAssetID)
-	if err != nil {
-		return respPacking(INVALID_PARAMS, err.Error())
+	var offset float64
+	if _, ok := params["offset"]; ok {
+		bucket, ok = params["offset"].(float64)
```
yilunzhang: this should be `offset, ok = ...`

api/common/interfaces.go
```diff
@@ -647,91 +647,152 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
...
+		if !ok {
+			return respPacking(INVALID_PARAMS, "offset should be a float64")
+		}
 	}
 
-	name, symbol, totalSupply, precision, err := chain.DefaultLedger.Store.GetAsset(assetID)
+	var limit float64
+	if _, ok := params["limit"]; ok {
+		bucket, ok = params["limit"].(float64)
```
yilunzhang: this should be `limit, ok = ...`

api/common/interfaces.go
```diff
@@ -647,91 +647,152 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
...
-		"precision":   precision,
+	subscribers, err := chain.DefaultLedger.Store.GetSubscribers(topic, uint32(bucket), uint32(offset), uint32(limit))
+	if err != nil {
+		return respPacking(INTERNAL_ERROR, err.Error())
 	}
-
-	return respPacking(SUCCESS, ret)
+	return respPacking(SUCCESS, map[string][]string{
+		"subscribers":         subscribers,
+		"subscribersInTxPool": localNode.GetTxnPool().GetSubscribers(),
```
yilunzhang: We should return subscribers in txpool only for the specified topic. In the current network, all subscribers in pool constantly contains ~30k results (30M in size)...

api/common/interfaces.go
trueinsider: when you unmarshal json into `map[string]interface{}` all numbers become float64. I can check that it falls within uint32 range or even into buckets count limit, it's just how uint32 is treated all over `interfaces.go` and I don't know if it's really necessary. Ideally every method should have their own request/response struct so that it'll be unmarshalled correctly or not unmarshalled at all.

yes, bucket are kept because there will still be some subscriptions in other buckets when we roll it out. but we should remove them from here sometimes in the future (when all non-zero bucket subscriptions are expired and new ones can't be created anymore)

api/common/interfaces.go
trueinsider: actually, no, because then you won't be able to access `offset` outside of that `if`

api/common/interfaces.go
trueinsider: same as above

api/common/interfaces.go
trueinsider: oh, you're right, I missed that somehow

trueinsider: also prob should be made optional

api/common/interfaces.go
yilunzhang: I think you missed the point, you are modifying `bucket` variable (instead of `offset`) using `params["offset"]`, same for below

trueinsider: oh, I see, thanks

api/common/interfaces.go
yilunzhang: I see. BTW I don't think anyone is using nonzero bucket AFAIK. Seems that everyone is just hard coding bucket=0 now 😂

chain/db/pubsub.go
```diff
@@ -238,14 +277,14 @@ func (sdb *StateDB) getSubscribers(topic string, bucket uint32) (map[string]stri
 
 		subscriberString := address.MakeAddressString(ps.subscriber, ps.identifier)
 
-		subscribers[subscriberString] = ps.meta
+		subscribers = append(subscribers, subscriberString)
```
yilunzhang: We are fundamentally changing the API here. IMO it's better to have an API or an option that can get both subscribers and metadata like before. These info are already read into the memory here, so it will be much more efficient if we can just include it in the response. If I'm using subscribers to store service providers' info, the price and other info will be in the meta. If I want to make a choice, I definitely need to know the price of every provider. With the current new API, I have to call getSubscription() for each provider, which is not feasible at all if there are hundreds of providers.

chain/txValidator.go
```diff
@@ -167,13 +169,24 @@ func CheckTransactionPayload(txn *transaction.Transaction) error {
...
 			return errors.New("Identifier too long")
 		}
 		if len(pld.Meta) > config.MaxTxnSubMetaList.GetValueAtHeight(h) {
 			return errors.New("Meta too long")
 		}
+	case pb.UNSUBSCRIBE_TYPE:
+		pld := payload.(*pb.Unsubscribe)
+
+		topic := pld.Topic
+		match, err := regexp.MatchString("(^[A-Za-z][A-Za-z0-9-_.+]{2,254}$)", topic)
```
yilunzhang: Do we need to check topic for unsubscribe at all? It should be enough if we simply check whether an existing subscription exist right?

api/common/interfaces.go
trueinsider: I have height limit after which it won't accept non-zero buckets anymore. If at that point there's no non-zero buckets in statedb, then we can safely remove that.

chain/txValidator.go
```diff
@@ -341,13 +354,24 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
-		subscriptionCount := DefaultLedger.Store.GetSubscribersCount(pld.Topic, pld.Bucket)
-		if subscriptionCount >= transaction.SubscriptionsLimit {
-			return fmt.Errorf("subscribtion count to %s can't be more than %d", pld.Topic, subscriptionCount)
+		pld := payload.(*pb.Unsubscribe)
+		subscribed, err := DefaultLedger.Store.IsSubscribed(pld.Topic, 0, pld.Subscriber, pld.Identifier)
+		if err != nil {
+			return err
+		}
+		if !subscribed {
+			return fmt.Errorf("subscribtion to %s doesn't exist", pld.Topic)
```
yilunzhang: typo: subscribtion -> subscription

chain/txValidator.go
```diff
@@ -341,13 +354,24 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
 		if err != nil {
 			return err
 		}
-		if subscribed {
-			return fmt.Errorf("subscriber %s already subscribed to %s", address.MakeAddressString(pld.Subscriber, pld.Identifier), pld.Topic)
+		if !subscribed {
+			subscriptionCount := DefaultLedger.Store.GetSubscribersCount(pld.Topic, pld.Bucket)
+			if subscriptionCount >= transaction.SubscriptionsLimit {
+				return fmt.Errorf("subscribtion count to %s can't be more than %d", pld.Topic, subscriptionCount)
```
yilunzhang: typo: subscribtion -> subscription

chain/db/pubsub.go
trueinsider: Problem is that returning meta for all subscribers could be A LOT of data, since we don't limit max page size and meta limit is quite big too. It could be an option with no meta by default.

BTW for tuna I'm choosing random provider and go with him, if price is ok for me. I don't search for the lowest, because that way providers can control who can connect to them (to attack them somehow, for example). So, I prefer random here with max acceptable price set.

chain/txValidator.go
trueinsider: existing subscription is gonna be checked later, doing leveldb search
if we can discard it earlier and with less work I think it could be worth it
but prob should create some common function to check topic name for both sub and unsub

chain/txValidator.go
```diff
@@ -682,13 +754,26 @@ func (bvs *BlockValidationState) CleanSubmittedTransactions(txns []*transaction.
...
-				delete(bvs.subscriptionCount, topic)
+					if bvs.subscriptionCount[topic] == 0 {
+						delete(bvs.subscriptionCount, topic)
+					}
+				}
+			}
+		case pb.UNSUBSCRIBE_TYPE:
+			unsubscribePayload := payload.(*pb.Unsubscribe)
+			topic := unsubscribePayload.Topic
+			key := subscription{topic, 0, string(unsubscribePayload.Subscriber), unsubscribePayload.Identifier}
```
yilunzhang: Do you think it's a good idea to separate bvs.subscriptions and bvs.unsubscriptions ? I kind of feel that when someone send a sub and unsub txn at the same time, it's very easy to have bugs if we mix them up (but it's just my feeling)

pb/transaction.go
```diff
@@ -114,6 +114,14 @@ func (m *Subscribe) ToMap() map[string]interface{} {
 	}
 }
 
+func (m *Unsubscribe) ToMap() map[string]interface{} {
+	return map[string]interface{}{
+		"subscriber": common.BytesToUint160(m.Subscriber),
```
yilunzhang: common.BytesToUint160 has been changed to common.HexStr for subscribe, we should also change it here (after rebase)

chain/db/pubsub.go
yilunzhang: We can make it by default not returning meta. Another thing we can have is a configurable parameter specifying max response size for (probably any) rpc, and if response is larger than it, just return error.

Also for tuna, random provider can work as the first version, but later I think it will be better to combine provider IP + price to choose the one that is (approximately) closest and cheapest.

chain/txValidator.go
yilunzhang: Yeah agree

chain/txValidator.go
trueinsider: yeah, it's easy to have bugs, so it doesn't allow sending sub and unsub at the same time

chain/txValidator.go
RealJohnSmith: Speaking about valid names, I think `/` and `*` should be allowed in the name. Using that you can make structures like subscription: `games/spacecats/matchmaking`, `games/spacecats/leaderboards`, `games/spacecats/*`
What do you think?

pb/transaction.go
trueinsider: done

chain/txValidator.go
trueinsider: I think we should probably discuss this out of scope of this PR

util/config/config.go
```diff
@@ -95,6 +95,10 @@ var (
 		heights: []uint32{133400, 0},
 		values:  []int{1024, MaxInt},
 	}
+	BucketsLimit = HeightDependentInt{ // ChangePoin for txn payload Subscribe.Meta
+		heights: []uint32{133400, 0},
```
NoelBright: 133400 is a invalid block number.

api/common/interfaces.go
```diff
@@ -634,10 +634,10 @@ func getAddressByName(s Serverer, params map[string]interface{}) map[string]inte
...
 }
 
-// getSubscribers get subscribers by topic
-// params: {"topic":<topic>, "bucket":<bucket>}
+// getSubscription get subscription
+// params: {"topic":<topic>, "bucket":<bucket>, "subscriber":<subscriber>}
 // return: {"resultOrData":<result>|<error data>, "error":<errcode>}
-func getSubscribers(s Serverer, params map[string]interface{}) map[string]interface{} {
+func getSubscription(s Serverer, params map[string]interface{}) map[string]interface{} {
 	if len(params) < 2 {
```
NoelBright: should be "< 3".

util/config/config.go
trueinsider: yeah, need to change this to some future block height when merging

api/common/interfaces.go
```diff
@@ -647,91 +647,169 @@ func getSubscribers(s Serverer, params map[string]interface{}) map[string]interf
...
-// getFirstAvailableTopicBucket get free topic bucket
-// params: {"topic":<topic>}
+// getSubscribersCount get subscribers count by topic
+// params: {"topic":<topic>, "bucket":<bucket>}
 // return: {"resultOrData":<result>|<error data>, "error":<errcode>}
-func getFirstAvailableTopicBucket(s Serverer, params map[string]interface{}) map[string]interface{} {
+func getSubscribersCount(s Serverer, params map[string]interface{}) map[string]interface{} {
 	if len(params) < 1 {
-		return respPacking(INVALID_PARAMS, "length of params is less than 1")
+		return respPacking(INVALID_PARAMS, "length of params is less than 2")
```
NoelBright: It's conflicted with "len(params) < 1"

NoelBright:
looks good to me.

trueinsider:
great, can we deploy it to internal cluster asap for testing?

transaction/payload.go
```diff
@@ -8,8 +8,7 @@ import (
 )
 
 const (
-	SubscriptionsLimit      = 1000
-	BucketsLimit            = 1000
+	SubscriptionsLimit      = 100000
 	MaxSubscriptionDuration = 65535
```
yilunzhang: What about doubling MaxSubscriptionDuration? I feel that the current limitation is a bit too short based on a few feedbacks

trueinsider: generally I'm not against that, but I feel like they think about it being too short because you can't extend existing subscription and you need to wait for it to end first. since that isn't the case after this PR is merged, just extending it every once in a while should be sufficient.

current max duration is about ~18.2 days

chain/db/pubsub.go
```diff
@@ -211,6 +210,24 @@ func (sdb *StateDB) subscribe(topic string, bucket uint32, subscriber []byte, id
...
+	ps, err := sdb.getPubSub(id)
+	if err != nil {
+		return err
+	}
+
+	if !ps.Empty() {
+		if err := sdb.cancelPubSubCleanupAtHeight(ps.expiresAt, id); err != nil {
+			return err
+		}
+		sdb.pubSub.Store(id, nil)
```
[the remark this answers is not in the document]
trueinsider: fixed, thanks

yilunzhang:
Passed tests on internal cluster

PR #613, opened 2019-08-06
Add NumLowFeeTxnPerBlock config to control building block logic

yilunzhang:
This provides a way to prevent recent txn spamming attack

### DIFF
--- a/chain/mining.go
+++ b/chain/mining.go
@@ -46,6 +46,7 @@ func (bm *BuiltinMining) BuildBlock(ctx context.Context, height uint32, chordID 
 	txnHashList = append(txnHashList, coinbase.Hash())
 	totalTxsSize := coinbase.GetSize()
 	txCount := 1
+	lowFeeTxCount := 0
 
 	if winnerType == pb.TXN_SIGNER {
 		if _, err = DefaultLedger.Store.GetTransaction(winnerHash); err != nil {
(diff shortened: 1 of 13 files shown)


chain/mining.go
```diff
@@ -75,27 +76,35 @@ func (bm *BuiltinMining) BuildBlock(ctx context.Context, height uint32, chordID
...
+		if config.Parameters.MinTxnFee > 0 {
+			if txn.UnsignedTx.Fee == 0 {
+				if uint32(zeroFeeTxCount) >= config.Parameters.NumZeroFeeTxnPerBlock {
+					log.Warning("too many zero fee transactions")
+					break
+				}
+			} else {
+				if txn.UnsignedTx.Fee < int64(config.Parameters.MinTxnFee) {
+					log.Warning("transaction fee is too low")
+					txnCollection.Pop()
```
gdmmx: It should not collection.Pop() but should zeroFeeTxCount++ if zeroFeeTxCount not reach NumZeroFeeTxnPerBlock.

yilunzhang: It pop() because the txn will never be packed into the block by this miner.
zeroFeeTxCount should only increase if and only if a zero fee txn is packed into the block.
PS: this is the old logic and has not been changed

gdmmx: It is unreasonable a `non-zero but less MinTxnFee fee` was rejected by peek `a zero fee` preferentially.

yilunzhang: `Pop()` has to be called whenever `continue` is called, otherwise it will get stuck

gdmmx:
```
Fee > MinTxnFee   // Accepted
Fee < MinTxnFee   // Rejected
Fee == 0          // Accepted * NumZeroFeeTxnPerBlock
```
It is unfair for `Fee < MinTxnFee`

PR #614, opened 2019-08-06
Dynamically adjust sync RAM usage based on total RAM

yilunzhang:
Dynamically adjust sync RAM usage based on (10% of) total RAM if sync batch window size if not set in config.

### DIFF
--- a/util/config/config.go
+++ b/util/config/config.go
@@ -20,6 +20,7 @@ import (
 	"github.com/nknorg/nkn/crypto/ed25519/vrf"
 	"github.com/nknorg/nkn/crypto/util"
 	"github.com/nknorg/nnet/transport"
+	"github.com/pbnjay/memory"
 )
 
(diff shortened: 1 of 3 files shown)


util/config/config.go
```diff
@@ -175,6 +177,11 @@ type Configuration struct {
 }
 
 func Init() error {
+	syncBatchWindowSize := uint32(memory.TotalMemory()*MaxSyncMemoryUsagePercent/100/MaxBlockSize) / Parameters.SyncBlocksBatchSize
+	if syncBatchWindowSize > 0 {
+		Parameters.SyncBatchWindowSize = syncBatchWindowSize
```
gdmmx: It will be overwrote by config.json.

yilunzhang: That's intended, this only changes the default value, user can always override it

PR #617, opened 2019-08-06
Fix state db race condition and txn pool sorting order

yilunzhang:
This PR fixed 2 bugs that has large impact on the network recently.

### DIFF
--- a/chain/collector.go
+++ b/chain/collector.go
@@ -64,9 +64,9 @@ func (s sortTxnsByPriceSize) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 
 func (s sortTxnsByPriceSize) Less(i, j int) bool {
 	if s[i].UnsignedTx.Fee == s[j].UnsignedTx.Fee {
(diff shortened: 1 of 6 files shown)


chain/collector.go
```diff
@@ -64,9 +64,9 @@ func (s sortTxnsByPriceSize) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 
 func (s sortTxnsByPriceSize) Less(i, j int) bool {
 	if s[i].UnsignedTx.Fee == s[j].UnsignedTx.Fee {
-		return s[i].GetSize() < s[j].GetSize()
+		return s[i].GetSize() > s[j].GetSize()
```
gdmmx: Why `return true` when `i > j` in a Less function?
It should return `i<j` and don't `sort.Reverse` after `sortTxnsByPrice`

yilunzhang: Size and price always have different order, the less here is defined by less price, but if same price, then larger size is effectively less price

NoelBright:
looks good to me.

PR #619, opened 2019-08-07
Improve TxPool memory management and dropping logic

yilunzhang:
* Drop txn only if txn hash match to prevent race condition
* Limit txpool memory size and sync blocks memory size
* Implement dropping nanopay txn in txpool


NoelBright:
Need nanopay txns to be drop? The nanopay txn has a nano_pay_expiration. If dropping it, the payment  maybe go to wrong.

yilunzhang:
@NoelBright I think we need to drop it, otherwise attacker can crash any node very easily. It should be receiver's responsibility to send txn in advance to prevent not receiving token

NoelBright:
ok

PR #633, opened 2019-08-29
Adjust default txpool/sync memory size and LowFeeTxnSizePerBlock

yilunzhang:
* Adjust default txpool/sync memory size and LowFeeTxnSizePerBlock
    * Make default txpool memory usage dynamic
    * Increase default sync memory usage to accelerate block syncing
    * Reduce LowFeeTxnSizePerBlock to slower down chain size grow
* Move Consequential to independent package
* Pass context to GenerateStateRoot


util/config/config.go
```diff
@@ -66,9 +66,11 @@ const (
 )
 
 const (
-	defaultConfigFile          = "config.json"
-	defaultSyncMemoryPercent   = 10
-	defaultSyncBatchWindowSize = 64
+	defaultConfigFile             = "config.json"
+	defaultSyncMaxMemoryPercent   = 25
+	defaultSyncBatchWindowSize    = 64
+	defaultTxPoolMaxMemoryPercent = 0.4
```
gdmmx: 4MB TxPoolMaxMemory is too many headcount for txPool since minimal size transaction just only 170~180 Bytes per txn.
4MB/180B ≈23.3K txn in txPool. It is costly for each getLength of sync.Map.

yilunzhang: The current value is 32M :)

4MB should be fine if we can eliminate memory leak. Currently there is no operation that will traverse all txn in txn pool unless a node is the block proposer, so the number of txn in pool is not the bottleneck at all. Instead, RAM usage is the current problem

chain/blockchain.go
```diff
@@ -31,7 +32,7 @@ func NewBlockchainWithGenesisBlock(store ILedgerStore) (*Blockchain, error) {
 		return nil, err
 	}
 
-	root, err := store.GenerateStateRoot(genesisBlock, false, false)
+	root, err := store.GenerateStateRoot(context.Background(), genesisBlock, false, false)
```
NoelBright: how to trigger it?

yilunzhang: GenerateStateRoot will not timeout when persist or create genesis block, but will timeout during consensus at https://github.com/nknorg/nkn/blob/master/consensus/proposal.go#L131

NoelBright: ok, I see

NoelBright:
Looks good to me.

PR #634, opened 2019-08-29
Add crypto puzzle restriction to generate ID txn

yilunzhang:
This can greatly reduce generate ID txn spams.

This is an incompatible update and the changing height is yet to be determined. Do not merge before we are ready to upgrade the network.


common/uint256.go
```diff
@@ -10,15 +10,21 @@ import (
...
+	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+}
+
 func (u *Uint256) CompareTo(o Uint256) int {
 	x := u.ToArray()
 	y := o.ToArray()
 
-	for i := len(x) - 1; i >= 0; i-- {
+	for i := 0; i < len(x); i++ {
```
gdmmx: Uint256.CompareTo change from descending cmp to ascending cmp, means Uint256 store BigEndian changed to LittleEndian.
Uint160 should be changed also for consistency, and additional highlight srcCode's comments and API docs.

yilunzhang: Also changed Uint160

PR #635, opened 2019-08-29
Rejecting low priority transaction when txpool is full

yilunzhang:
This should greatly reduce node load (CPU, RAM, IO) when txpool is full.

### DIFF
--- a/chain/pool/txpool.go
+++ b/chain/pool/txpool.go
@@ -17,22 +17,28 @@ import (
 	"github.com/nknorg/nkn/util/log"
 )
 
+func compareTxnPriority(txn1, txn2 *transaction.Transaction) int {
+	if txn1.UnsignedTx.Fee > txn2.UnsignedTx.Fee {
+		return 1
+	}
+	if txn1.UnsignedTx.Fee < txn2.UnsignedTx.Fee {
+		return -1
+	}
+	if txn1.GetSize() > txn2.GetSize() {
+		return -1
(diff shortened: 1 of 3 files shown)


chain/pool/txpool.go
```diff
@@ -17,22 +17,28 @@ import (
...
+		return -1
+	}
+	if txn1.GetSize() > txn2.GetSize() {
+		return -1
 	}
-	return s[i].UnsignedTx.Fee < s[j].UnsignedTx.Fee
+	if txn1.GetSize() < txn2.GetSize() {
+		return 1
+	}
+	return 0
```
gdmmx: How about succinct style:
```
{
  delta := txn1.UnsignedTx.Fee - txn2.UnsignedTx.Fee
  if delta == 0 {
      return int32(txn2.GetSize() - txn1.GetSize())    // return Size: txn2 - txn1
  }
  return delta     // return Fee: txn1 - txn2
}
```

NoelBright:
Looks good to me.

chain/pool/txpool.go
yilunzhang: There are 2 reasons:
1. It's a standard to always return {+1, 0, -1} in cmp function. For example, some of our code use `if Compare(xxx, xxx) == 1` (not here though, you can search for `== 1` globally).
2. In theory subtraction of two uint32 can overflow int32 (although very rare)

PR #640, opened 2019-09-05
Add height as VerifyTransaction argument for Go SDK usage

yilunzhang:
Currently height will be read from local chain db, so sdk cannot use VerifyTransaction function to verify transaction.


NoelBright:
looks good to me

PR #644, opened 2019-09-15
Implement random sampling neighbors in voting and gossip

yilunzhang:
This mechanism greatly increases consensus convergence under balanced initial condition according to our simulations.


node/neighbor.go
```diff
@@ -129,3 +130,88 @@ func (localNode *LocalNode) startConnectingToRandomNeighbors() {
...
+
+	for _, pred := range c.Predecessors() {
+		chordNeighborIDs[string(pred.Id)] = struct{}{}
+	}
+
+	for _, fingers := range c.FingerTable() {
+		for _, finger := range fingers {
+			chordNeighborIDs[string(finger.Id)] = struct{}{}
+		}
+	}
```
gdmmx: It seems that we should unique(chordNeighborIDs) at here.

node/neighbor.go
```diff
@@ -129,3 +130,88 @@ func (localNode *LocalNode) startConnectingToRandomNeighbors() {
...
+
+	numRandomSamples := int(config.VotingSampleRandomNeighbor * float32(len(randomNeighbors)))
+	if numRandomSamples > 0 {
+		rand.Shuffle(len(randomNeighbors), func(i, j int) { randomNeighbors[i], randomNeighbors[j] = randomNeighbors[j], randomNeighbors[i] })
+		sampledNeighbors = append(sampledNeighbors, randomNeighbors[:numRandomSamples]...)
+	}
+
+	return sampledNeighbors
+}
```
gdmmx: How about extract an internal function getNeighbors(filter, chordNum, randomNum) and invoke it as below:
```
GetGossipNeighbors(filter) {
    getNeighbors(filter, config.GossipSampleChordNeighbor, config.GossipSampleRandomNeighbor)
}
GetVotingNeighbors(filter) {
    getNeighbors(filter, config.VotingSampleChordNeighbor, config.VotingSampleRandomNeighbor)
}
```

yilunzhang: Good idea. Pushed the new version.

PR #645, opened 2019-09-24
Add block check when syncing block

NoelBright:
add TransactionCheck and HeaderCheck to blockchain.AddBlock.

### DIFF
--- a/block/block.go
+++ b/block/block.go
@@ -16,8 +16,10 @@ import (
 )
 
 type Block struct {
-	Header       *Header
-	Transactions []*transaction.Transaction
+	Header          *Header
+	Transactions    []*transaction.Transaction
+	IsHeaderChecked bool
+	IsTxnsChecked   bool
 }
 
 func (b *Block) FromMsgBlock(msgBlock *pb.Block) {
--- a/chain/blockchain.go
+++ b/chain/blockchain.go
@@ -57,6 +57,7 @@ func (bc *Blockchain) AddBlock(block *block.Block, fastAdd bool) error {
 
 	err := bc.SaveBlock(block, fastAdd)
 	if err != nil {
+		log.Error("AddBlock error, ", err)
 		return err
 	}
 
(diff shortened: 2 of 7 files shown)


block/block.go
```diff
@@ -16,8 +16,10 @@ import (
 )
 
 type Block struct {
-	Header       *Header
-	Transactions []*transaction.Transaction
+	Header                *Header
+	Transactions          []*transaction.Transaction
+	IsHeaderChecked       bool
+	IsTransactionsChecked bool
```
gdmmx: How about rename `IsTransactionsChecked` as `IsTxnsChecked`?

chain/blockchain.go
```diff
@@ -57,6 +57,7 @@ func (bc *Blockchain) AddBlock(block *block.Block, fastAdd bool) error {
 
 	err := bc.SaveBlock(block, fastAdd)
 	if err != nil {
+		log.Error("AddBlock error, ", err)
```
gdmmx: Why `SaveBlock` error but log output "**AddBlock** error"?

chain/blockchain.go
```diff
@@ -72,6 +73,24 @@ func (bc *Blockchain) GetHeader(hash Uint256) (*block.Header, error) {
...
+			}
+			block.IsHeaderChecked = true
+		}
+
+		if !block.IsTransactionsChecked {
+			err := TransactionCheck(context.Background(), block)
+			if err != nil {
+				return err
+			}
+			block.IsTransactionsChecked = true
```
gdmmx: Same as above. Short name for srcCode concision.

chain/blockchain.go
NoelBright: because this statement is in AddBlock function.

block/block.go
NoelBright: done

chain/blockchain.go
NoelBright: done

PR #647, opened 2019-09-25
Implement multicast and compression protocol

yilunzhang:

### DIFF
--- a/api/websocket/server/relay.go
+++ b/api/websocket/server/relay.go
@@ -49,10 +49,30 @@ func (ws *WsServer) sendOutboundRelayMessage(srcAddrStrPtr *string, msg *pb.Outb
 		return
 	}
 
+	payloads := msg.Payloads
+	if len(payloads) == 0 && len(msg.Payload) > 0 {
+		payloads = append(payloads, msg.Payload)
+	}
+
+	if len(payloads) == 0 {
+		log.Warningf("no payload")
+		return
+	}
+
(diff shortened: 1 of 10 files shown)


pb/clientmessage.proto
```diff
@@ -34,6 +40,7 @@ message OutboundMessage {
   uint32 nonce = 5;
   bytes block_hash = 6;
   repeated bytes signatures = 7;
+  repeated bytes payloads = 8;
```
NoelBright: What is the difference between OutboundMessage.Payload and OutboundMessage.Payloads?

api/websocket/server/relay.go
```diff
@@ -49,10 +49,30 @@ func (ws *WsServer) sendOutboundRelayMessage(srcAddrStrPtr *string, msg *pb.Outb
 		return
 	}
 
+	payloads := msg.Payloads
```
NoelBright: Is "sendOutboundRelayMessage" called by client?

pb/clientmessage.proto
yilunzhang: Payloads is meant to replace Payload, similar to dests vs dest

api/websocket/server/relay.go
yilunzhang: It's called in `api/websocket/server/server.go`

NoelBright:
Looks good to me.

PR #664, opened 2019-12-12
Prevent RPC handler from panic before ledger init

yilunzhang:


yilunzhang:
> What happen if RPC request arrived before nnet.chord layer ready?

It happens when node is computing ID and it will return ErrNullID error in GetNetNode()

PR #671, opened 2019-12-30
Fix tls websocket address and certificate error

yilunzhang:
* Add root/intermediate certificate
* Use remote node wss port instead of local port
* Fix wss wrong node error contains ws addr 


node/localnode.go
```diff
@@ -400,9 +405,9 @@ func (localNode *LocalNode) findAddr(key []byte, tls bool) (string, []byte, []by
 	if tls == true {
 		if nodeData.TlsWebsocketDomain != "" {
 			host = nodeData.TlsWebsocketDomain
-			port = config.Parameters.HttpWssPort
+			port = uint16(nodeData.TlsWebsocketPort)
 		} else {
```
gdmmx: It seems that unnecessary to `address, err := url.Parse(pred.Addr) && host := address.Hostname()`([L394](https://github.com/nknorg/nkn/blob/2492f04d6985a4b4c2c3dc5a45e55f1e61c3b3fd/node/localnode.go#L394) ~ [L402](https://github.com/nknorg/nkn/blob/2492f04d6985a4b4c2c3dc5a45e55f1e61c3b3fd/node/localnode.go#L402))  if `tls != true`

yilunzhang: You are right, fixed.

api/websocket/server/server.go
```diff
@@ -146,13 +146,24 @@ func (ws *WsServer) registryMethod() {
...
+
+		isTlsClient := cmd["IsTls"].(bool)
+		var addr, localAddr string
+		var pubkey, id []byte
+
+		if isTlsClient {
+			addr, pubkey, id, err = localNode.FindWssAddr(clientID)
+			localAddr = localNode.GetWssAddr()
+		} else {
+			addr, pubkey, id, err = localNode.FindWsAddr(clientID)
```
gdmmx: Could we have others more distinctive API name for FindWsAddr & FindWs**s**Addr ? It is too similar to cause typo.

gdmmx: Same as GetWsAddr & localNode.GetWs**s**Addr

yilunzhang: It's for naming consistent. If we want to change it, we'd better change all the related naming project-wise in another PR

gdmmx: 👌

PR #672, opened 2020-01-07
Add new name registration function

bufrr:


chain/store/name.go
```diff
@@ -0,0 +1,276 @@
...
+func getNameId(name string) []byte {
+	nameId := sha256.Sum256([]byte(strings.ToLower(name)))
+	return nameId[:hashPrefixLength]
+}
+
+func (sdb *StateDB) getRegistrant(name string) ([]byte, uint32, error) {
+	nameId := getNameId(name)
+	ni, err := sdb.getNameInfo(nameId)
+	if err != nil {
+		return nil, 0, errors.New("name registration info not found")
```
gdmmx: Why override error info from low-level API?
It will hided the root cause of failure for up-level side.

chain/store/name.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+	}
+
+	enc, err := sdb.trie.TryGet(append(NamePrefix, nameId...))
+	if err != nil {
+		return nil, err
+	}
+	ni := new(nameInfo)
+	if len(enc) > 0 {
+		ni := new(nameInfo)
```
gdmmx: Double alloc `nameInfo` twice?

chain/blockValidator.go
```diff
@@ -130,7 +130,7 @@ func TransactionCheck(ctx context.Context, block *block.Block) error {
 			return fmt.Errorf("transaction block check failed: %v", err)
 		}
 		bvs.Commit()
-		if err := VerifyTransactionWithLedger(txn); err != nil {
+		if err := VerifyTransactionWithLedger(txn, block.Header.UnsignedHeader.Height); err != nil {
```
yilunzhang: We don't need to pass block height to `VerifyTransactionWithLedger` (if we need, we should pass ledger height, not block.height). We can get ledger height inside the function. Search for `DefaultLedger.Store.GetHeight()` in `VerifyTransactionWithLedger` and you will find it

chain/txValidator.go
```diff
@@ -265,7 +271,7 @@ func CheckTransactionPayload(txn *transaction.Transaction, height uint32) error
 }
 
 // VerifyTransactionWithLedger verifys a transaction with history transaction in ledger
-func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
+func VerifyTransactionWithLedger(txn *transaction.Transaction, height uint32) error {
```
yilunzhang: We should remove line `height := DefaultLedger.Store.GetHeight()` in VerifyTransactionWithLedger because height is now passed in as argument

chain/store/stateGenerator.go
```diff
@@ -57,29 +57,28 @@ func (cs *ChainStore) spendTransaction(states *StateDB, txn *transaction.Transac
 			return err
 		}
 
-		if err := states.UpdateBalance(pg[0], config.NKNAssetID, Fixed64(txn.UnsignedTx.Fee), Subtraction); err != nil {
-			return err
-		}
 		states.IncrNonce(pg[0])
```
yilunzhang: It's better to increment nonce after update balance

chain/store/stateGenerator.go
```diff
@@ -57,29 +57,28 @@ func (cs *ChainStore) spendTransaction(states *StateDB, txn *transaction.Transac
...
 
 		registerNamePayload := pl.(*pb.RegisterName)
 		if config.LegacyNameService.GetValueAtHeight(height) {
 			states.setName_legacy(registerNamePayload.Registrant, registerNamePayload.Name)
+			if err := states.UpdateBalance(pg[0], config.NKNAssetID, Fixed64(txn.UnsignedTx.Fee), Subtraction); err != nil {
```
yilunzhang: We can use the same code for legacy name and new name service because `RegistrationFee` will always be zero for legacy transaction. So the logic could be:
1. update balance
2. increment nonce
3. if old height, then setName_legacy, otherwise registerName

util/config/config.go
```diff
@@ -97,6 +99,14 @@ var (
 		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
 	}
+	AllowNameRegex = HeightDependentString{
+		heights: []uint32{7500, 0},
+		values:  []string{"(^[A-Za-z][A-Za-z0-9-_.+]{5,64}$)", "(^[A-Za-z][A-Za-z0-9-_.+]{2,254}$)"},
```
yilunzhang: 64 -> 63
64 char is the same as public key and should be avoided

chain/store/name.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func (sdb *StateDB) cancelNameCleanupAtHeight(height uint32, name string) error {
+	ids, err := sdb.getPubSubCleanup(height)
+	if err != nil {
+		return err
+	}
+	if _, ok := ids[name]; ok {
+		delete(ids, name)
+	}
```
gdmmx: It is unnecessary for `comma ok` before `delete`. Just `delete(ids, name)` is enough.

[Reference](https://golang.org/src/builtin/builtin.go?h=delete#L146)
```
// The delete built-in function deletes the element with the specified key
// (m[key]) from the map. If m is nil or there is no such element, delete
// is a no-op.
func delete(m map[Type]Type1, key Type)
```

chain/store/stateGenerator.go
bufrr: There are some differences between old and new name service when update balance, because old name service doesnt have registration fee.

api/common/interfaces.go
```diff
@@ -668,6 +628,27 @@ func getSubscription(s Serverer, params map[string]interface{}) map[string]inter
...
+	}
+
+	registrant, expiresAt, err := chain.DefaultLedger.Store.GetRegistrant(name)
+	if err != nil {
+		return respPacking(INTERNAL_ERROR, err.Error())
+	}
+	reg := hex.EncodeToString(registrant)
+	response := make(map[string]interface{})
+	response["registrant"] = reg
+	response["expiresAt"] = expiresAt
```
gdmmx: Suggestion:
```
response := map[string]interface{}{
        "registrant": reg,
        "expiresAt": expiresAt,
    }
```
Integrated allocation and assignment is friendly with compiler optimization.

chain/store/name.go
```diff
@@ -68,7 +67,7 @@ func (sdb *StateDB) getRegistrant(name string) ([]byte, uint32, error) {
 	nameId := getNameId(name)
 	ni, err := sdb.getNameInfo(nameId)
 	if err != nil {
-		return nil, 0, errors.New("name registration info not found")
+		return nil, 0, fmt.Errorf("name registration info not found, %v", err)
```
yilunzhang: We can simply return err here because `getNameInfo` error is not always not found error

chain/store/stateGenerator.go
```diff
@@ -57,29 +57,28 @@ func (cs *ChainStore) spendTransaction(states *StateDB, txn *transaction.Transac
...
-			}
-			donationAddress, err := ToScriptHash(config.DonationAddress)
-			if err != nil {
-				return err
-			}
-			if err = states.UpdateBalance(donationAddress, config.NKNAssetID, Fixed64(registerNamePayload.RegistrationFee), Addition); err != nil {
-				return err
-			}
 		}
+		states.IncrNonce(pg[0])
```
yilunzhang: This line should be removed because `IncrNonce` is already called above

chain/store/stateGenerator.go
```diff
@@ -57,13 +57,40 @@ func (cs *ChainStore) spendTransaction(states *StateDB, txn *transaction.Transac
...
+		if err = states.UpdateBalance(pg[0], config.NKNAssetID, Fixed64(registerNamePayload.RegistrationFee)+Fixed64(txn.UnsignedTx.Fee), Subtraction); err != nil {
+			return err
+		}
+		donationAddress, err := ToScriptHash(config.DonationAddress)
+		if err != nil {
+			return err
+		}
+		if err = states.UpdateBalance(donationAddress, config.NKNAssetID, Fixed64(registerNamePayload.RegistrationFee), Addition); err != nil {
+			return err
+		}
```
yilunzhang: We are missing one thing: miner should get the txn fee

chain/txValidator.go
```diff
@@ -321,36 +331,83 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
+			pk, err := crypto.DecodePoint(pld.Registrant)
+			if err != nil {
+				return err
+			}
+			addrHash, err := program.CreateProgramHash(pk)
+			if err != nil {
+				return err
+			}
+			balance := DefaultLedger.Store.GetBalance(addrHash)
+			if int64(balance) < pld.RegistrationFee {
```
yilunzhang: Should be reg fee + txn fee

chain/store/name.go
```diff
@@ -139,6 +141,11 @@ func (sdb *StateDB) deleteNameInfo(nameId string) error {
 }
 
 func (sdb *StateDB) FinalizeNames(commit bool) {
+	_, height, _ := sdb.cs.getCurrentBlockHashFromDB()
+	if config.LegacyNameService.GetValueAtHeight(height) {
```
yilunzhang: We should use `height+1` because new txn will be in block height+1

chain/txValidator.go
```diff
@@ -161,6 +161,7 @@ func CheckTransactionPayload(txn *transaction.Transaction, height uint32) error
 		if !match {
 			fmt.Errorf("name should match regex %s", regexPattern)
 		}
+	case pb.TRANSFER_NAME_TYPE:
```
yilunzhang: We should validate `registrant` field of register, transfer, delete name (check length)

chain/txValidator.go
```diff
@@ -321,36 +331,83 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction) error {
...
+			} else if name != pld.Name {
+				return fmt.Errorf("no name %s registered for pubKey %+v", pld.Name, pld.Registrant)
+			}
+		} else {
+			registrant, _, err := DefaultLedger.Store.GetRegistrant(pld.Name)
+			if err != nil {
+				return err
+			}
+			if len(registrant) == 0 {
+				fmt.Errorf("name doesn't exist")
```
yilunzhang: missing return

chain/txValidator.go
```diff
@@ -387,10 +387,17 @@ func VerifyTransactionWithLedger(txn *transaction.Transaction, height uint32) er
 		if err != nil {
 			return err
 		}
+		if len(registrant) == 0 {
+			return fmt.Errorf("can not transfer unregistered name")
+		}
 		if bytes.Equal(registrant, pld.Recipient) {
 			return fmt.Errorf("can not transfer names to its owner")
 		}
-		if !bytes.Equal(registrant, pld.Registrant) {
```
yilunzhang: Are we missing `pld.Registrant` verification here?

bufrr: fixed

PR #674, opened 2020-01-14
Remove invalid txn if CleanSubmittedTransactions fails

yilunzhang:
This PR fixes a bug where txn pool consists of invalid txn due to the race condition of block generation and new txn entering pool. This PR introduces a mechanism to remove invalid txn from pool.


chain/pool/txpool.go
```diff
@@ -587,9 +589,24 @@ func (tp *TxnPool) deleteTransactionFromMap(txn *transaction.Transaction) {
 }
 
 func (tp *TxnPool) CleanBlockValidationState(txns []*transaction.Transaction) error {
```
bufrr: I think this function wont return any error except nil, right?

yilunzhang: That's true, but in the future we might add error return to this function

PR #677, opened 2020-02-18
Fixed node not syncing if there is only one persist node in neighbors

yilunzhang:
This PR fixes a bug where node cannot start sync if there is only one persist finished node in its neighbors. Also, this PR fixes the camel field name in pb message field. It will not change the generated Go type.


pb/transaction.proto
```diff
@@ -58,7 +58,7 @@ message Coinbase {
 }
 
 message SigChainTxn {
-	bytes sigChain     = 1;
```
bufrr: Will it cause data incompatibility, if some nodes serialize by the old definition?

yilunzhang: It won't because protobuf convert snake_case to camelCase automatically for go. Also, even we change the field name completely, the data will still be compatible because protobuf only use the field id (`= 1` part) for serialization/deserialization.

PR #679, opened 2020-03-03
Simplify docker build and run workflow

yilunzhang:
* Add web directory to git repo so build will no longer need nodejs and yarn
* Add docker entrypoint to copy default config and certs, and create wallet if not exists in data volume


docker/entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
+
+if echo "$1" | grep -q "^nknd"; then
+  ! [ -s web ] && echo "Copying web directory..." && cp -R ../web web
+  ! [ -s certs ] && echo "Copying default certs..." && cp -R ../certs certs
+  ! [ -s config.json ] && echo "Copying default config.json..." && cp /nkn/config.mainnet.json config.json
+  ! [ -s wallet.json ] && ! [ -s wallet.pswd ] && echo "Creating wallet.pswd..." && head -c 1024 /dev/urandom | tr -dc A-Za-z0-9 | head -c 32 > wallet.pswd && echo >> wallet.pswd
```
gdmmx: auto-create wallet.json & wallet.pswd will conflict with `--web-gui-create-wallet` feature

docker/Dockerfile
```diff
@@ -12,6 +12,9 @@ COPY --from=builder /nkn/nknd /nkn/
 COPY --from=builder /nkn/nknc /nkn/
 COPY --from=builder /nkn/config.mainnet.json /nkn/
 COPY --from=builder /nkn/certs /nkn/certs
+COPY --from=builder /nkn/web /nkn/web
+COPY --from=builder /nkn/docker/entrypoint.sh /nkn/
 RUN ln -s /nkn/nknd /nkn/nknc /usr/local/bin/
 WORKDIR /nkn/data
-CMD nknd
+ENTRYPOINT ["/nkn/entrypoint.sh"]
+CMD ["nknd", "--password-file", "wallet.pswd", "--web-gui-listen-address", "0.0.0.0"]
```
gdmmx: Has webUI listen on `0.0.0.0` enough security?

docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG base
-FROM golang:1.12.4-alpine as builder
+FROM golang:1.13.8-alpine as builder
```
bufrr: Maybe we can use golang 1.14, which is the newest golang version?

docker/Dockerfile
yilunzhang: Yes docker has its own port map rule

docker/Dockerfile
yilunzhang: We haven't throughly tested the node with 1.14 yet. We can probably do more testing before upgrade the version.

docker/entrypoint.sh
yilunzhang: Fixed

gdmmx:
LGTM

PR #681, opened 2020-03-10
Handle receiving blocks with invalid signature

yilunzhang:
* Node will not cache block with invalid signature
* Add neighbor who sent block with invalid signature to blacklist

### DIFF
--- a/chain/blockValidator.go
+++ b/chain/blockValidator.go
@@ -309,6 +309,11 @@ func SignerCheck(header *block.Header) error {
 		return fmt.Errorf("invalid block signer public key %x, should be %x", header.UnsignedHeader.SignerPk, publicKey)
 	}
 
+	_, err = crypto.DecodePoint(header.UnsignedHeader.SignerPk)
+	if err != nil {
+		return fmt.Errorf("decode public key error: %v", err)
+	}
+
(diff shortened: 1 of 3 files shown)


chain/blockValidator.go
```diff
@@ -309,6 +309,11 @@ func SignerCheck(header *block.Header) error {
 		return fmt.Errorf("invalid block signer public key %x, should be %x", header.UnsignedHeader.SignerPk, publicKey)
 	}
 
+	_, err = crypto.DecodePoint(header.UnsignedHeader.SignerPk)
```
bufrr: Can we remote this check?PubKey length check looks useless here.

yilunzhang: This is actually checking that `header.UnsignedHeader.SignerPk` is a valid hex string

yilunzhang:
> Double check `header.UnsignedHeader.SignerPk` validity in `SignerCheck` & `SignatureCheck`.
> Is there any scene invoke `SignerCheck` individual without `SignatureCheck`? Otherwise we check `UnsignedHeader.SignerPk` once is enough.

Block passing `SignerCheck` but not `SignatureCheck` will be cached. We need to make sure block with invalid pk to be cached, otherwise an attacker can continue sending it in the future and node will continue to fetch it.

PR #682, opened 2020-03-11
Add log.Fatal as replacement for panic

yilunzhang:
* Add fatal log level
* Replace panic with log.Fatal

This PR is based on #680, will rebase once #680 is merged.


yilunzhang:
> The API of `sdb.FinalizeXXX` were changed. What happen if one of it return `EmptyUint256`?
> Should we roll back the sdb.trie if `sdb.FinalizeXXX` return err in halfway?

I don't think we need rollback here. If `Finalize()` returns error, commit will not be called and db will not be changed.

PR #683, opened 2020-03-13
Remove old log files if total log file size exceeds a limit

yilunzhang:
This PR is based on #682, will rebase once #682 is merged.

### DIFF
--- a/util/config/config.go
+++ b/util/config/config.go
@@ -174,6 +174,7 @@ var (
 		MiningDebug:                  true,
 		LogLevel:                     1,
 		MaxLogFileSize:               20,
+		MaxLogFileTotalSize:          100,
 		SyncBatchWindowSize:          0,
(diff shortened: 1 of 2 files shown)


chain/store/name.go
```diff
@@ -168,39 +168,50 @@ func (sdb *StateDB) deleteNameInfo(nameId string) error {
...
+			if err != nil {
+				return false
 			}
 			if commit {
 				sdb.namesCleanup.Delete(height)
 			}
 		}
 		return true
 	})
+	return err
```
bufrr: Maybe `return nil` here is more clear, as err can only be `nil`.

yilunzhang: `err` can be non nil. Return at L206 is not for `FinalizeNames` but for `sdb.namesCleanup.Range` iterator.

PR #685, opened 2020-03-17
Use periodical ping/pong message to detect client timeout rapidly

yilunzhang:
Use periodical ping/pong message to detect client timeout rapidly. This can prevent reduce message lost caused by sending to broken clients.

### DIFF
--- a/api/websocket/server/relay.go
+++ b/api/websocket/server/relay.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/nknorg/nkn/pb"
@@ -65,7 +66,7 @@ func (ws *WsServer) sendOutboundRelayMessage(srcAddrStrPtr *string, msg *pb.Outb
 func (ws *WsServer) sendInboundMessage(clientID string, inboundMsg *pb.InboundMessage) bool {
 	clients := ws.SessionList.GetSessionsById(clientID)
 	if clients == nil {
-		log.Infof("Client Not Online: %s", clientID)
+		log.Debugf("Client Not Online: %s", clientID)
 		return false
 	}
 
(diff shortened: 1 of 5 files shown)


api/websocket/server/server.go
```diff
@@ -41,6 +41,7 @@ const (
 	pingInterval                 = 8 * time.Second
 	pongTimeout                  = 10 * time.Second // should be greater than pingInterval
```
bufrr: I think it's better to set  both `pongTimeout`  and `pingInterval` as Duration type, although they may have the same value.

yilunzhang: What do you mean? They are already `time.Duration` type now

bufrr: My bad, I misunderstood

api/websocket/server/relay.go
```diff
@@ -124,11 +125,44 @@ func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay) {
 				sigChainLen: int(relayMessage.SigChainLen),
 			})
 		}
-	} else {
+		if time.Duration(relayMessage.MaxHoldingSeconds) > pongTimeout/time.Second {
+			ok := ws.messageDeliveredCache.Push(relayMessage)
+			if !ok {
+				log.Warningf("MessageDeliveredCache full, discarding messages.")
+			}
+		}
```
gdmmx: Why still cache msg even sendInboundMessage success?

api/websocket/server/relay.go
```diff
@@ -124,11 +125,44 @@ func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay) {
...
+	} else if relayMessage.MaxHoldingSeconds > 0 {
 		ws.messageBuffer.AddMessage(clientID, relayMessage)
 	}
 }
 
+func (ws *WsServer) startCheckingLostMessages() {
+	for {
```
gdmmx: Should we need something interval for each loop in order to avoid eat all cpu.idle?

api/websocket/server/relay.go
```diff
@@ -124,11 +125,44 @@ func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay) {
...
+		v, ok := ws.messageDeliveredCache.Pop()
```
gdmmx: Where to push again if the popped msg still can't deliver?

api/websocket/server/relay.go
```diff
@@ -124,11 +125,44 @@ func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay) {
...
+				success := false
+				for _, client := range clients {
+					if client.GetLastReadTime().After(threshold) {
+						success = true
+						break
+					}
+				}
+				if success {
+					continue
+				}
```
gdmmx: What happen if a multi-dest msg had received by part of clients?
This piece looks that the msg will be marked **success** and skip to next pop() if one of dest client `GetLastReadTime` after `threshold `.

api/websocket/server/relay.go
yilunzhang: `sendInboundMessage` does not mean the client will definitely receive message when there is a proxy between node and client.

api/websocket/server/relay.go
yilunzhang: It won't. Pop will block until the internal channel is available: https://github.com/nknorg/nkn/pull/685/files#diff-6bfb9c74eef98d9279404e3f0e69f0ffR39

api/websocket/server/relay.go
yilunzhang: When msg from messageBuffer failed to deliver, it will be put back into buffer again

api/websocket/server/relay.go
yilunzhang: There is no such thing called multi-dest msg, each msg only has one dest. A multi-dest msg will be separated into multiple msgs.

api/websocket/server/relay.go
gdmmx: I know Pop will block if channel is empty. I means that the popped msg still can't deliver and will be pushed back again very quickly. The channel is not "empty" but pop/push same msg repeatedly with high frequency.

yilunzhang: Popped msg that can't deliver will be put back into message buffer, but messageDeliveredCache, so it won't happen

PR #689, opened 2020-03-31
Improve transaction fee handle and verification logic

yilunzhang:
* Unify fee subtract and nonce increment logic
* Check amount and fee range when checking payload

### DIFF
--- a/chain/store/stateGenerator.go
+++ b/chain/store/stateGenerator.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/nknorg/nkn/block"
 	"github.com/nknorg/nkn/chain"
-	. "github.com/nknorg/nkn/common"
+	"github.com/nknorg/nkn/common"
 	"github.com/nknorg/nkn/crypto"
 	"github.com/nknorg/nkn/pb"
 	"github.com/nknorg/nkn/program"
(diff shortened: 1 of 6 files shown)


chain/txValidator.go
```diff
@@ -89,10 +88,6 @@ func CheckTransactionAttribute(txn *transaction.Transaction) error {
 	return nil
 }
 
-func checkAmountPrecise(amount Fixed64, precision byte) bool {
-	return amount.GetData()%int64(math.Pow(10, 8-float64(precision))) != 0
-}
-
```
gdmmx: Why get rid of precise check?

yilunzhang: Because the precise check is doing nothing (checking int % 1 == 0)... You can verify that...

chain/store/stateGenerator.go
```diff
@@ -7,239 +7,187 @@ import (
...
 		return err
 	}
 
+	pg, err := txn.GetProgramHashes()
+	if err != nil {
+		return err
+	}
+
+	if txn.UnsignedTx.Fee > 0 {
+		if err = states.UpdateBalance(pg[0], config.NKNAssetID, common.Fixed64(txn.UnsignedTx.Fee), Subtraction); err != nil {
```
gdmmx: Here reimplement to balance subtract `Tx.Fee` & `payload.Amount` individual.
Could statesDB revert to previous state if sub `payload.Amount` met err later?

yilunzhang: It won't. DB is changed in batch. Before .Commit() is called, no changes will be persisted

gdmmx: That is OK.

chain/txValidator.go
gdmmx: It is meaningful for others asset which precision < 8.

PR #692, opened 2020-04-02
Remove legacy crypto big int code from old ECDSA signature

yilunzhang:


chain/blockValidator.go
```diff
@@ -309,9 +309,9 @@ func SignerCheck(header *block.Header) error {
 		return fmt.Errorf("invalid block signer public key %x, should be %x", header.UnsignedHeader.SignerPk, publicKey)
 	}
 
-	_, err = crypto.DecodePoint(header.UnsignedHeader.SignerPk)
+	err = crypto.CheckPublicKey(header.UnsignedHeader.SignerPk)
```
bufrr, line 312: I think it will be more specific to rename it like `CheckPubLicKeySize/Length`

chain/mining.go
```diff
@@ -174,7 +174,7 @@ func (bm *BuiltinMining) BuildBlock(ctx context.Context, height uint32, chordID
 				TransactionsRoot: txnRoot.ToArray(),
 				WinnerHash:       winnerHash.ToArray(),
 				WinnerType:       winnerType,
-				SignerPk:         bm.account.PublicKey.EncodePoint(),
+				SignerPk:         bm.account.PublicKey,
```
bufrr, line 177: Will this change cause hard fork?

crypto/ed25519/ed25519.go
```diff
@@ -16,50 +16,33 @@ const PrivateKeySize = ed25519.PrivateKeySize
...
-	return privKey, X, Y, nil
+func GenKeyPair() ([]byte, []byte, error) {
+	return ed25519.GenerateKey(rand.Reader)
 }
 
-func Sign(priKey []byte, data []byte) (*big.Int, *big.Int, error) {
-	sig := ed25519.Sign(priKey, data)
-	return new(big.Int).SetBytes(sig[:ed25519.SignatureSize/2]), new(big.Int).SetBytes(sig[ed25519.SignatureSize/2:]), nil
+func Sign(privateKey, data []byte) ([]byte, error) {
+	if len(privateKey) != PrivateKeySize {
```
bufrr, line 24: I think `ed25519.Sign` will check privatekey length itself?

chain/mining.go
yilunzhang: It won't, the previous `PublicKey.EncodePoint()` is the same as `PublicKey` now, both are the bytes of pk

chain/blockValidator.go
yilunzhang: Actually I was planning to put more pubkey check into that function if we have in the future, but I don't have anything in mind now.

node/node.go
```diff
@@ -15,7 +15,7 @@ import (
 type Node struct {
 	*nnetpb.Node
 	*pb.NodeData
-	publicKey *crypto.PubKey
+	publicKey []byte
```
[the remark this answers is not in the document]
yilunzhang: Actually not, they are the same. The reason we have crypto.PubKey was for ECDSA key pair (which was encoded by X and Y), but for Ed25519, only a single bytes (X) is used.

crypto/ed25519/ed25519.go
yilunzhang: `ed25519.Sign` will panic if length is incorrect, and we don't want panic here.

PR #694, opened 2020-04-03
Add minimal number of neighbor samples to help bootstrap network

yilunzhang:
This solves the insufficient propagation problem when there are only a few nodes and helps network to bootstrap faster.

### DIFF
--- a/node/neighbor.go
+++ b/node/neighbor.go
@@ -1,7 +1,6 @@
 package node
 
 import (
-	"math"
 	"math/rand"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


node/neighbor.go
```diff
@@ -179,17 +178,29 @@ func (localNode *LocalNode) splitNeighbors(filter func(*RemoteNode) bool) ([]*Re
...
+func (localNode *LocalNode) getSampledNeighbors(filter func(*RemoteNode) bool, chordNeighborSampleRate float64, chordNeighborMinSample int, randomNeighborSampleRate float64, randomNeighborMinSample int) []*RemoteNode {
 	var sampledNeighbors []*RemoteNode
 	chordNeighbors, randomNeighbors := localNode.splitNeighbors(filter)
 
-	numChordSamples := int(math.Ceil(chordNeighborSampleRate * float64(len(chordNeighbors))))
+	numChordSamples := int(chordNeighborSampleRate * float64(len(chordNeighbors)))
+	if numChordSamples < chordNeighborMinSample {
+		numChordSamples = chordNeighborMinSample
+	}
+	if numChordSamples > len(chordNeighbors) {
```
bufrr, line 189: Can `numChordSamples > len(chordNeighbors)` and `numChordSamples < chordNeighborMinSample` happen together?

yilunzhang: Yes, if `len(chordNeighbors) < chordNeighborMinSample `, both of these `if` conditions could be true

PR #699, opened 2020-04-11
Add rate limiting for RPC requests and sync block msg

yilunzhang:


node/localnode.go
```diff
@@ -188,6 +168,31 @@ func NewLocalNode(wallet vault.Wallet, nn *nnet.NNet) (*LocalNode, error) {
 	return localNode, nil
 }
 
+func (localNode *LocalNode) MarshalJSON() ([]byte, error) {
```
bufrr, line 171: Why moving this function from L51, any difference?

yilunzhang: It's the same, but just a convention to put receiver function after create function (NewXXX)

node/syncblock.go
```diff
@@ -192,6 +196,10 @@ func (localNode *LocalNode) getBlocksMessageHandler(remoteMessage *RemoteMessage
 		return replyBuf, false, nil
 	}
 
+	if !localNode.syncBlockLimiter.AllowN(time.Now(), int(endHeight-startHeight)) {
```
bufrr, line 199: If node received x block(x<endHeight-startHeight), it will return false, is it correct?

yilunzhang: Actually not, this is a rate limiter: https://pkg.go.dev/golang.org/x/time/rate?tab=doc It only allows other nodes to request N blocks per second on average, optionally with a burst

api/httpjson/RPCserver.go
```diff
@@ -47,22 +41,26 @@ type ServeMux struct {
...
-
-	return server
 }
 
-// this is the funciton that should be called in order to answer an rpc call
+// Handle is the funciton that should be called in order to answer an rpc call
 // should be registered like "http.HandleFunc("/", httpjsonrpc.Handle)"
 func (s *RPCServer) Handle(w http.ResponseWriter, r *http.Request) {
+	if !s.limiter.Allow() {
+		w.WriteHeader(http.StatusTooManyRequests)
```
bufrr, line 60: Will this affect legal node's connection when illegal node exists?

yilunzhang: It might be if node is under DDoS attack, but a node cannot distinguish between honest node and attacker node, so he can only reject all of them

PR #701, opened 2020-04-17
Add http json rpc address in getws[s]addr response

yilunzhang:
This enables client to use node's http json rpc address so it no longer relies on seed node once connected.


api/common/interfaces.go
```diff
@@ -361,9 +361,10 @@ func getVersion(s Serverer, params map[string]interface{}) map[string]interface{
 	return respPacking(SUCCESS, config.Version)
 }
 
-func NodeInfo(addr string, pubkey, id []byte) map[string]string {
+func NodeInfo(wsAddr, rpcAddr string, pubkey, id []byte) map[string]string {
 	nodeInfo := make(map[string]string)
-	nodeInfo["addr"] = addr
+	nodeInfo["addr"] = wsAddr
```
bufrr, line 366: Maybe `wsAddr` sounds more specific?

yilunzhang: This will break existing clients

yilunzhang:
> Why we need to return rpcAddr when client call `getwsaddr`?

The purpose is to help client not relying on seed node once they join the network. With the rpc address returned in results, clients can now send transactions and make rpc requests to connected node first, and using seed node only as a backup.

PR #705, opened 2020-04-30
Upgrade wallet format: use scrypt as kdf instead of sha256

yilunzhang:
Upgrade wallet format: use scrypt as kdf instead of sha256. Also removed redundant fields in wallet.json.

Wallet version is now 2. Version 1 will continue to work, but less secure.

Use `nknc wallet --changepassword` to upgrade to v2 wallet.

### DIFF
--- a/vault/data.go
+++ b/vault/data.go
@@ -40,7 +40,7 @@ type WalletData struct {
 	Scrypt        *ScryptData
 }
 
-func NewWalletData(account *Account, password, masterKey, iv []byte) (*WalletData, error) {
+func NewWalletData(account *Account, password, masterKey, iv, scryptSalt []byte, scryptN, scryptR, scryptP int) (*WalletData, error) {
 	if len(masterKey) == 0 {
 		masterKey = util.RandomBytes(MasterKeyLength)
 	}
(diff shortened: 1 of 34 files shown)


vault/wallet.go
```diff
@@ -256,84 +125,74 @@ func (w *WalletImpl) Sign(txn *transaction.Transaction) error {
...
-	// check original password
-	oldPasswordKey := crypto.ToAesKey(oldPassword)
-	passwordKeyHash, err := HexStringToBytes(w.Data.PasswordHash)
+func (w *Wallet) ChangePassword(oldPassword, newPassword []byte) error {
+	account, err := w.DecryptAccount(oldPassword)
+	if err != nil {
+		return err
+	}
+
+	address, err := account.ProgramHash.ToAddress()
```
bufrr, line 138: Maybe accout here didn't decrypted correctly, failed to change password with an old version of wallet file when using new nknc client

yilunzhang: The only possibility that wallet decrypt failed is because the wallet file is modified and corrupted, and we can do nothing about it.

vault/data.go
```diff
@@ -0,0 +1,184 @@
...
+	common.ClearBytes(passwordhash[:])
+	return passwordhash2[:]
+}
+
+func PasswordToAesKeyScrypt(password []byte, scryptData *ScryptData) ([]byte, error) {
+	salt, err := hex.DecodeString(scryptData.Salt)
+	if err != nil {
+		return nil, err
+	}
+	return scrypt.Key(password, salt, scryptData.N, scryptData.R, scryptData.P, 32)
```
bufrr, line 199: maybe we should use `MasterKeyLength` here instead of 32?

yilunzhang: 32 is actually not `MasterKeyLength`, but AES key length, which cannot be changed (just like sha256 output is 32 bytes)

PR #707, opened 2020-05-05
Reduce sigchain cache timeout to reduce relay ram usage (stage 2/2)

yilunzhang:
This PR is the second stage in reducing relay RAM usage when most nodes have merged #702


yilunzhang:
> So drastic fluctuation of reduce expired time from 10min to 10sec.
> SigChain only be cached half of consensus duration. Has it enough time for propagate sufficiently?

In last version (stage 1) we introduced a new protocol `PinSigChain`, that will extend potentially useful sigchains' cache duration to 10 min.

PR #727, opened 2020-07-21
Lets encrypt certificate integration

bufrr:


yilunzhang:
A lot of dependencies are added to `go.sum`. Did you run `go mod tidy` before committing?

api/common/common.go
```diff
@@ -34,3 +38,12 @@ func ResponsePack(errCode ErrCode) map[string]interface{} {
 	}
 	return resp
 }
+
+func ReturnCert(h *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
yilunzhang, line 42: Suggest rename to `GetCert` or `GetCertificate`

api/websocket/server/server.go
```diff
@@ -487,19 +487,8 @@ func (ws *WsServer) Broadcast(data []byte) error {
...
-	// load cert
-	cert, err := tls.LoadX509KeyPair(CertPath, KeyPath)
-	if err != nil {
-		log.Error("load keys fail", err)
-		return nil, err
-	}
-
 	tlsConfig := &tls.Config{
-		Certificates: []tls.Certificate{cert},
+		GetCertificate: api.ReturnCert,
```
yilunzhang, line 491: Previously wss cert (HttpWssCert) might be different from https json cert (HttpsJsonCert), but now HttpWssCert is no longer being used

certs/cert.go
```diff
@@ -0,0 +1,75 @@
+package certs
+
+import (
+	"os"
+	"time"
+
+	"github.com/nknorg/nkn/v2/util/config"
+
```
yilunzhang, line 8: Did you run gofmt?

certs/cert.go
```diff
@@ -0,0 +1,75 @@
...
+				if err != nil {
+					log.Errorf("apply cert failed: %v", err)
+					break
+				}
+				err = saveCertToDisk(cert)
+				if err != nil {
+					log.Errorf("save cert failed: %v", err)
+					break
+				}
+			} else if cm.NeedApply() {
```
yilunzhang, line 59: Should this be `NeedRenew`?

certs/cert.go
```diff
@@ -0,0 +1,75 @@
...
+		CertStableURL:     c.CertStableURL,
+		PrivateKey:        c.PrivateKey,
+		Certificate:       c.Certificate,
+		IssuerCertificate: c.IssuerCertificate,
+		CSR:               c.CSR,
+	}
+}
+
+func PrepareCert() {
+	for {
```
yilunzhang, line 41: Why 2 levels for loop? 1 loop should be enough with proper logic

certs/cert.go
```diff
@@ -0,0 +1,75 @@
...
+		for {
+			cm, err := NewCertManager()
```
yilunzhang, line 43: Why create a new cm each time?

certs/user.go
```diff
@@ -0,0 +1,74 @@
...
+	b, err := ioutil.ReadFile(config.Parameters.ACMEUserFile)
+	if err == nil {
+		// user exists
+		err = json.Unmarshal(b, &u)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// get domain name
+		tlsDomain := &node.TlsDomain{DashedIP: node.DotToDash(config.Parameters.Hostname)}
```
yilunzhang, line 57: `config.Parameters.Hostname` is empty be default

certs/user.go
```diff
@@ -0,0 +1,74 @@
...
+			return nil, err
+		}
+		u.Domain = commonName
+		// create private Key
+		privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+		if err != nil {
+			return nil, err
+		}
+
+		u.Email = commonName + "@nkn.org" // fake email address
```
yilunzhang, line 69: Is domain and email field required? If not we can omit them

certs/user.go
```diff
@@ -0,0 +1,74 @@
...
+func (u *User) GetEmail() string {
+	return u.Email
+}
+func (u User) GetRegistration() *registration.Resource {
+	return u.Registration
+}
+func (u *User) GetPrivateKey() crypto.PrivateKey {
+	return u.Key
+}
+func (u *User) SaveToDisk() error {
```
yilunzhang, line 33: Why do we want to save it to disk? node IP might change at any time, so it's better to get the latest one each time nknd starts, not read the previous one

certs/certManager.go
```diff
@@ -0,0 +1,173 @@
...
+			return nil, err
+		}
+	} else {
+		log.Info("read resourceFile err: %v", err)
+	}
+
+	return &certManager{User: u, LegoConfig: c, LegoClient: client, Resource: cache.toResource()}, nil
+}
+
+func (cm *certManager) ApplyNewCert() (*certificate.Resource, error) {
```
yilunzhang, line 70: Suggest rename to `ObtainNewCert`

certs/certManager.go
```diff
@@ -0,0 +1,173 @@
...
+	log.Infof("[%s] acme: %d hours remaining, renewBefore: %d\n", cm.Resource.Domain, int(timeLeft.Hours()), config.Parameters.CertRenewBefore)
+
+	cert, err := cm.LegoClient.Certificate.Renew(*cm.Resource, true, false)
+	if err != nil {
+		return nil, err
+	}
+	return cert, nil
+}
+
+func (cm *certManager) NeedApply() bool {
```
yilunzhang, line 111: Suggest rename to `NeedObtain`

certs/certManager.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// write certificate resource to disk
+	err = ioutil.WriteFile(config.Parameters.ACMEResourceFile, b, certFileMode)
+	if err != nil {
+		return err
+	}
+
+	// write certificate PEM to disk
+	err = ioutil.WriteFile(config.Parameters.HttpsJsonCert, cert.Certificate, certFileMode)
```
yilunzhang, line 162: The cert we obtained should not overwrite user provided cert

yilunzhang:
Two logics are missing:
1. User provided cert should be used first. We should only try to obtain cert if user does not specify a valid cert/key
2. https/wss listeners should start only after cert is read/obtained/renewed successfully

api/common/common.go
```diff
@@ -34,3 +38,12 @@ func ResponsePack(errCode ErrCode) map[string]interface{} {
...
+	cert, err := tls.LoadX509KeyPair(config.Parameters.HttpsJsonCert, config.Parameters.HttpsJsonKey)
```
gdmmx, line 43: How about load KeyPair specified by tls.ClientHelloInfo.ServerName in order to support SNI (RFC 4366)?

api/httpjson/RPCserver.go
```diff
@@ -312,3 +312,26 @@ func (s *RPCServer) SetLocalNode(ln *node.LocalNode) {
 func (s *RPCServer) GetNetNode() *node.LocalNode {
 	return s.GetLocalNode()
 }
+
+func (s *RPCServer) setRPCTlsInfo() error {
```
yilunzhang, line 316: It's better to set local node when creating local node because local node info might be cached by neighbors once created and client might get outdated info.

api/websocket/server/server.go
```diff
@@ -237,7 +240,7 @@ func (ws *WsServer) Restart() {
 		time.Sleep(time.Second)
 		ws.Stop()
 		time.Sleep(time.Second)
-		go ws.Start()
+		go ws.Start(nil)
```
yilunzhang, line 243: nil channel will never fire any event, so wss ready will never be fired

certs/cert.go
```diff
@@ -0,0 +1,176 @@
...
+		CertStableURL:     c.CertStableURL,
+		PrivateKey:        c.PrivateKey,
+		Certificate:       c.Certificate,
+		IssuerCertificate: c.IssuerCertificate,
+		CSR:               c.CSR,
+	}
+}
+
+func VerifyOCSP(bundle []byte) (bool, error) {
+	privateKey, _ := rsa.GenerateKey(rand.Reader, 2048)
```
yilunzhang, line 48: Better to handle error

certs/cert.go
```diff
@@ -0,0 +1,176 @@
...
+				continue
+			}
+			err = saveCertToDisk(renew)
+			if err != nil {
+				log.Errorf("save renew cert failed: %v", err)
+				continue
+			}
+		}
+	}
+	return
```
yilunzhang, line 162: No need to return at the end of function

certs/user.go
```diff
@@ -0,0 +1,72 @@
...
+		// create private Key
+		privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+		if err != nil {
+			return nil, err
+		}
+		u.Key = privateKey
+	}
+	// get domain name
+	tlsDomain := &node.TlsDomain{DashedIP: node.DotToDash(config.Parameters.Hostname)}
+	commonName, err := tlsDomain.IpToDomain(config.Parameters.DefaultTlsDomainTmpl)
```
yilunzhang, line 65: What will happen if user is read from disk but domain is changed? Will let's encrypt handle this case correctly?

yilunzhang: And do you want to use `GetDefaultDomainFromIP` instead?

certs/cert.go
```diff
@@ -0,0 +1,176 @@
...
+				continue
+			}
+			err = saveCertToDisk(cert)
+			if err != nil {
+				log.Errorf("save cert failed: %v", err)
+				continue
+			}
+			if !wssCertExists {
+				changeToDefault(false, !wssCertExists)
+				wssCertReady <- struct{}{}
```
yilunzhang, line 143: You need to add default select to avoid blocking after first loop

certs/cert.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			if !httpsCertExists {
+				changeToDefault(!httpsCertExists, false)
+				httpsCertReady <- struct{}{}
```
yilunzhang, line 147: Same here

cmd/nknd/nknd.go
```diff
@@ -228,6 +214,41 @@ func nknMain(c *cli.Context) error {
 		config.Parameters.Hostname = extIP
 	}
 
+	wssCertExists, httpsCertExists, _ := certs.CertsExists()
+	wssCertReady := make(chan struct{})
+	httpsCertReady := make(chan struct{})
```
yilunzhang, line 219: Better to make it length 1 instead of 0

common/common.go
```diff
@@ -105,8 +105,11 @@ func SliceRemove(slice []uint32, h uint32) []uint32 {
 }
 
 func FileExisted(filename string) bool {
-	_, err := os.Stat(filename)
-	return err == nil || os.IsExist(err)
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
```
yilunzhang, line 112: If `filename` is a directory, this function will return false, and the caller will attempt to create the file with the same name and will fail.

util/config/config.go
```diff
@@ -207,12 +207,13 @@ var (
...
-		HttpWssCert:                  "certs/STAR.ipv4.nknlabs.io.cert",
-		HttpWssKey:                   "certs/STAR.ipv4.nknlabs.io.key",
-		HttpsJsonDomain:              "{{.DashedIP}}.ipv4.nknlabs.io",
-		HttpsJsonCert:                "certs/STAR.ipv4.nknlabs.io.cert",
-		HttpsJsonKey:                 "certs/STAR.ipv4.nknlabs.io.key",
+		DefaultTlsDomainTmpl:         "{{.DashedIP}}.ipv4.nknlabs.io",
+		DefaultTlsCert:               "certs/nknd.cert",
+		DefaultTlsKey:                "certs/nknd.key",
+		ACMEUserFile:                 "certs/user.json",
+		ACMEResourceFile:             "certs/resource.json",
```
yilunzhang, line 214: It's better to include domain in these filenames: `DefaultTlsCert`, `DefaultTlsKey`, `ACMEUserFile`, `ACMEResourceFile`. This way nknd don't need to re-obtain certificate if IP is changed from A to B and then back to A.

api/websocket/server/server.go
bufrr: this function were unused, should we delete it?

api/httpjson/RPCserver.go
bufrr: we can not set it correctly before we obtain the certs, and we obtain certs functions were execution asynchronous

api/websocket/server/server.go
yilunzhang: I think we can delete it

api/httpjson/RPCserver.go
```diff
@@ -292,9 +272,35 @@ func (s *RPCServer) Start() {
...
+		timer := time.NewTimer(time.Second * 5)
+		select {
+		case <-httpsCertReady:
+			log.Info("https cert ready")
+			tlsListener, err := s.initTlsListen()
+			if err != nil {
+				log.Errorf("Https Cert: %v", err.Error())
+				return
+			}
+			//s.setRPCTlsInfo()
```
yilunzhang, line 298: Remove unused code

api/httpjson/RPCserver.go
```diff
@@ -292,9 +272,35 @@ func (s *RPCServer) Start() {
...
+			s.httpServer.Serve(tlsListener)
+		case <-timer.C:
+			log.Info("https server is now unavailable")
```
yilunzhang, line 301: Suggest change to "https server is unavailable yet"

api/httpjson/RPCserver.go
```diff
@@ -292,9 +272,35 @@ func (s *RPCServer) Start() {
...
+		case <-httpsCertReady:
+			log.Info("https cert ready")
+			tlsListener, err := s.initTlsListen()
+			if err != nil {
+				log.Errorf("Https Cert: %v", err.Error())
+				return
+			}
+			//s.setRPCTlsInfo()
+			s.httpServer.Serve(tlsListener)
+		case <-timer.C:
```
yilunzhang, line 300: You don't have to change this but you can use `<-time.After(5*time.Second)` to be more succinct.

api/httpjson/RPCserver.go
```diff
@@ -292,9 +272,35 @@ func (s *RPCServer) Start() {
...
+		select {
+		case <-httpsCertReady:
+			log.Info("https cert ready")
+			tlsListener, err := s.initTlsListen()
+			if err != nil {
+				log.Errorf("Https Cert: %v", err.Error())
+				return
+			}
+			//s.setRPCTlsInfo()
+			s.httpServer.Serve(tlsListener)
```
yilunzhang, line 299: You need to capture returned error, print it and return.

api/websocket/server/server.go
```diff
@@ -109,8 +104,23 @@ func (ws *WsServer) Start() error {
 	ws.server = &http.Server{Handler: http.HandlerFunc(ws.websocketHandler)}
 	go ws.server.Serve(ws.listener)
 
-	ws.tlsServer = &http.Server{Handler: http.HandlerFunc(ws.websocketHandler)}
-	go ws.tlsServer.Serve(ws.tlsListener)
+	go func(wssCertReady chan struct{}) {
```
yilunzhang, line 107: All comments for http tls server applies to here as well

api/websocket/server/server.go
```diff
@@ -109,8 +104,23 @@ func (ws *WsServer) Start() error {
...
+		for {
+			timer := time.NewTimer(time.Second * 5)
+			select {
+			case <-wssCertReady:
+				log.Info("wss cert ready")
+				ws.tlsListener, err = ws.initTlsListen()
+				if err != nil {
+					log.Error("Https Cert: ", err.Error())
+				}
+				ws.tlsServer = &http.Server{Handler: http.HandlerFunc(ws.websocketHandler)}
```
yilunzhang, line 117: Why websocket separate `ws.server` and `ws.tlsServer` while http only has a single `s.httpServer`?

certs/cert.go
```diff
@@ -0,0 +1,183 @@
...
+		log.Error(err)
+	}
+	prePreparedCert := len(config.Parameters.HttpsJsonDomain) > 0 && len(config.Parameters.HttpWssDomain) > 0
+	defaultDomain, err := node.GetDefaultDomainFromIP(config.Parameters.Hostname, config.Parameters.DefaultTlsDomainTmpl)
+	if err != nil {
+		log.Error(err)
+	}
+	userFile, resourceFile, certFile, keyFile := GetACMEFileNames(defaultDomain, config.Parameters.CertPath)
+	defaultCertExists, err := VerifyLocalCertificate(certFile, keyFile, defaultDomain)
+	config.Parameters.DefaultTlsCert = certFile
```
yilunzhang, line 117: It's better to move L117-L121 to a separate function. `CertExists()` sounds like a function that only check if cert exists or not, but actually it's also changing the config, which is not expected

yilunzhang: Another way is to move L217-L233 in nknd.go into a single function call that returns two channel: wssCertReady and httpsCertReady

cmd/nknd/nknd.go
```diff
@@ -197,12 +199,46 @@ func nknMain(c *cli.Context) error {
...
+			extIP, err = AskMyIP(config.Parameters.SeedList)
+		}
+		if err != nil {
+			return err
+		}
+		log.Infof("My IP address is %s", extIP)
+		config.Parameters.Hostname = "64.64.245.149"
+	}
+
+	wssCertExists, httpsCertExists, prePreparedCert, _ := certs.CertExists()
```
yilunzhang, line 217: It's better to move L217-L233 into a single function call that returns two channel: wssCertReady and httpsCertReady

cmd/nknd/nknd.go
```diff
@@ -197,12 +199,46 @@ func nknMain(c *cli.Context) error {
...
+		if createMode { // There is no seed available in create mode, used ipify
+			extIP, err = ipify.GetIp()
+		} else {
+			extIP, err = AskMyIP(config.Parameters.SeedList)
+		}
+		if err != nil {
+			return err
+		}
+		log.Infof("My IP address is %s", extIP)
+		config.Parameters.Hostname = "64.64.245.149"
```
yilunzhang, line 214: You forgot to change this

cmd/nknd/nknd.go
```diff
@@ -278,6 +299,7 @@ func nknMain(c *cli.Context) error {
 
 	// set JsonRPC server localnode
 	rpcServer.SetLocalNode(localNode)
+	go rpcServer.StartTls(httpsCertReady)
```
yilunzhang, line 302: Why not call it together with `rpcServer.StartHttp()`?

cmd/nknd/nknd.go
```diff
@@ -197,12 +199,46 @@ func nknMain(c *cli.Context) error {
...
+		go certs.PrepareCert(wssCertExists, httpsCertExists, wssCertReady, httpsCertReady)
+	}
+
 	// init web service
 	dashboard.Init(nil, wallet, nil)
 
 	// start JsonRPC
 	rpcServer := httpjson.NewServer(nil, wallet)
-	rpcServer.Start()
+	rpcServer.Init()
```
yilunzhang, line 240: Suggest to move Init, StartHttp, StartTls into a single function call Start() just like websocket part

util/config/config.go
```diff
@@ -249,6 +247,14 @@ type Configuration struct {
...
 	HttpsJsonKey                 string        `json:"HttpsJsonKey"`
 	HttpsJsonPort                uint16        `json:"HttpsJsonPort"`
+	DefaultTlsCert               string        `json:"DefaultTlsCert"`
+	DefaultTlsKey                string        `json:"DefaultTlsKey"`
+	DefaultTlsDomainTmpl         string        `json:"DefaultTlsDomainTmpl"`
+	ACMEUserFile                 string        `json:"ACMEUserFile"`
+	ACMEResourceFile             string        `json:"ACMEResourceFile"`
+	CertRenewBefore              uint16        `json:"CertRenewBefore"`   //in hours
+	CertCheckInterval            time.Duration `json:"CertCheckInterval"` // in seconds
+	CertPath                     string        `json:"CertPath"`
```
yilunzhang, line 257: Suggest rename to "CertDirectory`

api/httpjson/RPCserver.go
```diff
@@ -292,9 +272,30 @@ func (s *RPCServer) Start() {
...
+		for {
+			select {
+			case <-httpsCertReady:
+				log.Info("https cert received")
+				tlsListener, err := s.initTlsListen()
+				if err != nil {
+					log.Errorf("Https Cert: %v", err.Error())
+					return
+				}
+				s.httpServer.Serve(tlsListener)
```
yilunzhang, line 293: Why not return after `Serve()`? Go routine will be leaked if serve returns error because `httpsCertReady` will not be fired again. Same for wss

certs/cert.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		log.Error(err)
+	}
+	prePreparedCert := len(config.Parameters.HttpsJsonDomain) > 0 && len(config.Parameters.HttpWssDomain) > 0
+	defaultDomain, err := node.GetDefaultDomainFromIP(config.Parameters.Hostname, config.Parameters.DefaultTlsDomainTmpl)
+	if err != nil {
+		log.Error(err)
+	}
+	_, _, certFile, keyFile := GetACMEFileNames(defaultDomain, config.Parameters.CertDirectory)
+	defaultCertExists, err := VerifyLocalCertificate(certFile, keyFile, defaultDomain)
```
yilunzhang, line 116: `err` should be printed (or omitted)

cmd/nknd/nknd.go
```diff
@@ -379,6 +383,28 @@ func netVersion(timer *time.Timer) {
 	}
 }
 
+func checkCerts() (chan struct{}, chan struct{}) {
```
yilunzhang, line 386: It's better to move this function to certs

cmd/nknd/nknd.go
```diff
@@ -379,6 +383,28 @@ func netVersion(timer *time.Timer) {
...
+		go func() {
+			wssCertReady <- struct{}{}
+		}()
+	}
+	if httpsCertExists {
+		go func() {
+			httpsCertReady <- struct{}{}
+		}()
+	}
+	if !(wssCertExists && httpsCertExists) || !prePreparedCert {
```
yilunzhang, line 402: `!(wssCertExists && httpsCertExists && prePreparedCert)` should be more clear

certs/cert.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		log.Error(err)
+	}
+
+	userFile, resourceFile, certFile, keyFile := GetACMEFileNames(defaultDomain, config.Parameters.CertDirectory)
+	config.Parameters.DefaultTlsCert = certFile
+	config.Parameters.DefaultTlsKey = keyFile
+	config.Parameters.ACMEUserFile = userFile
+	config.Parameters.ACMEResourceFile = resourceFile
+	if !https || !prePreparedCert {
```
yilunzhang, line 183: I think the logic is not right. Assume user provides a https certificate but not wss certificate, this if condition should be false, but it will be true currently

certs/cert.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	prePreparedCert := len(config.Parameters.HttpsJsonDomain) > 0 && len(config.Parameters.HttpWssDomain) > 0
+	defaultDomain, err := node.GetDefaultDomainFromIP(config.Parameters.Hostname, config.Parameters.DefaultTlsDomainTmpl)
+	if err != nil {
+		log.Error(err)
+	}
+	_, _, certFile, keyFile := GetACMEFileNames(defaultDomain, config.Parameters.CertDirectory)
+	defaultCertExists, err := VerifyLocalCertificate(certFile, keyFile, defaultDomain)
+
+	return wssUserCertExists || defaultCertExists, httpsUserCertExists || defaultCertExists, prePreparedCert
```
yilunzhang, line 118: I think return `wssUserCertExists, httpsUserCertExists, defaultCertExists` should be easier

certs/certManager.go
```diff
@@ -0,0 +1,187 @@
...
+	keySuffix := ".key"
+	userFile := filepath.Join(dirName, domainName+".user.json")
+	resourceFile := filepath.Join(dirName, domainName+".resource.json")
+	certFile := filepath.Join(dirName, domainName+certSuffix)
+	keyFile := filepath.Join(dirName, domainName+keySuffix)
+	return userFile, resourceFile, certFile, keyFile
+}
+
+func saveCertToDisk(cert *certificate.Resource) error {
+	// JSON encode certificate resource
```
yilunzhang, line 153: Should create `config.Parameters.CertDirectory` directory if not exists

certs/user.go
```diff
@@ -0,0 +1,72 @@
...
+func (u *User) GetEmail() string {
+	return u.Email
+}
+func (u User) GetRegistration() *registration.Resource {
+	return u.Registration
+}
+func (u *User) GetPrivateKey() crypto.PrivateKey {
+	return u.Key
+}
+func (u *User) SaveToDisk() error {
```
yilunzhang, line 33: Should create `config.Parameters.CertDirectory` directory if not exists

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	if certs[0].Subject.CommonName != domain {
+		return false, fmt.Errorf("inconsistent domain names")
+	}
+	return true, nil
+}
+
+func VerifyLocalCertificate(certPath string, keyPath string, domain string) (bool, error) {
+	if len(certPath) > 0 && len(keyPath) > 0 && len(domain) > 0 && common.FileExisted(certPath) && common.FileExisted(keyPath) {
+		return VerifyCertificate(certPath, domain)
```
yilunzhang, line 97: Should we also verify key is valid?

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+		go func() {
+			wssCertReady <- struct{}{}
+		}()
+	}
+	if httpsCertExists || defaultCertExists {
+		go func() {
+			httpsCertReady <- struct{}{}
+		}()
+	}
+	if !(wssCertExists && httpsCertExists) && !defaultCertExists {
```
yilunzhang, line 141: Should this just be `!(wssCertExists && httpsCertExists)`? If default cert exists, we still needs to renew certificate right?

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+	_, _, certFile, keyFile := GetACMEFileNames(defaultDomain, config.Parameters.CertDirectory)
+	defaultCertExists, err := VerifyLocalCertificate(certFile, keyFile, defaultDomain)
+	if err != nil {
+		log.Error(err)
+	}
+
+	return wssUserCertExists, httpsUserCertExists, defaultCertExists
+}
+
+func CheckCerts() (chan struct{}, chan struct{}) {
```
yilunzhang, line 124: Suggest to rename to PrepareCerts, and rename the current PrepareCerts to prepareCerts

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			err = saveCertToDisk(cert)
+			if err != nil {
+				log.Errorf("save cert failed: %v", err)
+				continue
+			}
+			if !wssCertExists {
+				select {
+				case wssCertReady <- struct{}{}:
+					log.Info("wss cert ready")
```
yilunzhang, line 173: You are missing default

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+				}
+			}
+			if !httpsCertExists {
+				select {
+				case httpsCertReady <- struct{}{}:
```
yilunzhang, line 178: You are missing default

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func ChangeToDefault(https, wss bool) {
+	defaultDomain, err := node.GetDefaultDomainFromIP(config.Parameters.Hostname, config.Parameters.DefaultTlsDomainTmpl)
+	if err != nil {
+		log.Error(err)
+	}
+
+	userFile, resourceFile, certFile, keyFile := GetACMEFileNames(defaultDomain, config.Parameters.CertDirectory)
+	config.Parameters.DefaultTlsCert = certFile
```
yilunzhang, line 204: Suggest to add an info log when config is changed:
Set default cert domain to <defaultDomain>

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+	config.Parameters.DefaultTlsKey = keyFile
+	config.Parameters.ACMEUserFile = userFile
+	config.Parameters.ACMEResourceFile = resourceFile
+	if !https {
+		config.Parameters.HttpsJsonCert = config.Parameters.DefaultTlsCert
```
yilunzhang, line 209: Suggest to add an info log when config is changed:
Use default https cert

api/certs/cert.go
```diff
@@ -0,0 +1,218 @@
...
+		config.Parameters.HttpsJsonKey = config.Parameters.DefaultTlsKey
+		config.Parameters.HttpsJsonDomain = config.Parameters.DefaultTlsDomainTmpl
+	}
+	if !wss {
+		config.Parameters.HttpWssCert = config.Parameters.DefaultTlsCert
```
yilunzhang, line 214: Suggest to add an info log when config is changed:
Use default wss cert

api/certs/certManager.go
```diff
@@ -0,0 +1,187 @@
...
+		CertStableURL:     cert.CertStableURL,
+		PrivateKey:        cert.PrivateKey,
+		Certificate:       cert.Certificate,
+		IssuerCertificate: cert.IssuerCertificate,
+		CSR:               cert.CSR,
+	}, "", "  ")
+	if err != nil {
+		return err
+	}
+	_ = os.Mkdir(config.Parameters.CertDirectory, certDirMode)
```
yilunzhang, line 168: Better to use `os.MkdirAll`

api/certs/user.go
```diff
@@ -0,0 +1,74 @@
...
+}
+func (u *User) GetPrivateKey() crypto.PrivateKey {
+	return u.Key
+}
+func (u *User) SaveToDisk() error {
+	b, err := json.MarshalIndent(u, "", "  ")
+	if err != nil {
+		return err
+	}
+	_ = os.Mkdir(config.Parameters.CertDirectory, certDirMode)
```
yilunzhang, line 39: Better to use `os.MkdirAll`

api/certs/cert.go
```diff
@@ -0,0 +1,230 @@
...
+			if err != nil {
+				log.Errorf("save cert failed: %v", err)
+				continue
+			}
+			if !wssCertExists {
+				select {
+				case wssCertReady <- struct{}{}:
+					log.Info("wss cert ready")
+				default:
+					log.Info("no wss cert available yet")
```
yilunzhang, line 180: This log is not right because cert is available. It should be something like wss cert channel full

api/certs/cert.go
```diff
@@ -0,0 +1,230 @@
...
+				}
+			}
+			if !httpsCertExists {
+				select {
+				case httpsCertReady <- struct{}{}:
+					log.Info("https cert ready")
+				default:
+					log.Info("no https cert available yet")
```
yilunzhang, line 188: Same here

PR #738, opened 2020-09-14
Prune ledger during first syncing to reduce syncing disk usage

yilunzhang:
This will greatly reduce the max disk usage during first syncing

### DIFF
--- a/util/config/config.go
+++ b/util/config/config.go
@@ -75,7 +75,6 @@ const (
 	MaxClientMessageSize           = 4 * 1024 * 1024
 	MinNameRegistrationFee         = 10 * common.StorageFactor
 	DefaultNameDuration            = 365 * 24 * 60 * 60 / int(ConsensusDuration/time.Second)
-	LivePruning                    = false
 )
 
 const (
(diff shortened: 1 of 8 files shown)


util/config/config.go
```diff
@@ -87,6 +86,8 @@ const (
 )
 
 var (
+	SyncPruning      = true
+	LivePruning      = false
```
bufrr, line 90: So that `LivePruning` option no longer configurable?

util/config/config.go
```diff
@@ -224,7 +225,7 @@ var (
 		PasswordFile:                 "",
 		StatePruningMode:             "lowmem",
 		RecentStateCount:             1024,
-		MinPruningCompactHeights:     4096,
+		MinPruningCompactHeights:     32768,
```
bufrr, line 228: Should we test it on $5 machine which have 30k uncompact block to see if it have enough space?

util/config/config.go
yilunzhang: It was not configurable before. We will only make it configurable if we are sure it will not cause any problem

util/config/config.go
yilunzhang: This parameter will have almost no effect on the disk usage because compact only happens when nknd starts. Typically nknd start very infrequently, e.g. once per month

bufrr (approved):

PR #739, opened 2020-09-15
Fix incorrect current block height after rollback

yilunzhang:

### DIFF
--- a/node/rollback.go
+++ b/node/rollback.go
@@ -1,6 +1,7 @@
 package node
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


node/rollback.go
```diff
@@ -20,29 +21,30 @@ const (
...
-		if config.MaxRollbackBlocks == 0 {
-			return false, err
-		}
-		log.Warning(err)
-	} else {
+	if majorityBlockHash == currentHash {
 		return false, nil
 	}
 
+	if config.MaxRollbackBlocks == 0 {
```
bufrr, line 39: I guess `config.MaxRollbackBlocks == 0` will always be false.

yilunzhang: That's true for now, but we have changed MaxRollbackBlocks before, it's just not configurable by users

bufrr (approved):

PR #740, opened 2020-09-15
Sync header in batch to reduce RAM usage during syncing

yilunzhang:

### DIFF
--- a/node/syncblock.go
+++ b/node/syncblock.go
@@ -302,6 +302,61 @@ func (remoteNode *RemoteNode) GetBlocks(startHeight, endHeight uint32) ([]*block
 	return blocks, nil
 }
 
+// getNeighborsBlockHeaderByHeight returns the block header at a given height
+// from given neighbors by calling GetBlockHeaders on all of them concurrently.
+func (localNode *LocalNode) getNeighborsBlockHeaderByHeight(height uint32, neighbors []*RemoteNode) (*sync.Map, error) {
+	var allHeaders sync.Map
+	var wg sync.WaitGroup
+	for _, neighbor := range neighbors {
+		wg.Add(1)
+		go func(neighbor *RemoteNode) {
+			defer wg.Done()
+			headers, err := neighbor.GetBlockHeaders(height, height)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


util/config/config.go
```diff
@@ -78,11 +78,14 @@ const (
 )
 
 const (
-	defaultConfigFile             = "config.json"
-	defaultSyncMaxMemoryPercent   = 25
-	defaultSyncBatchWindowSize    = 64
-	defaultTxPoolMaxMemoryPercent = 0.4
-	defaultTxPoolMaxMemorySize    = 32
+	defaultConfigFile                 = "config.json"
+	estimatedHeaderSize               = 418
```
gdmmx, line 82: What is magic number of 418?
Suggest use named const variate or `unsafe.Sizeof(blk.Header)`

yilunzhang: It's measured from current db. It cannot be get from size because size of bytes cannot be determined from structure definition (can be any length). When node first starts without ledger, he only has genesis block, which has different header size.

node/syncblock.go
```diff
@@ -302,6 +302,61 @@ func (remoteNode *RemoteNode) GetBlocks(startHeight, endHeight uint32) ([]*block
...
+	}
+	wg.Wait()
+	return &allHeaders, nil
+}
+
+// getNeighborsMajorityBlockHashByHeight returns the majority of given
+// neighbors' block hash at a give height
+func (localNode *LocalNode) getNeighborsMajorityBlockHashByHeight(height uint32, neighbors []*RemoteNode) common.Uint256 {
+	for i := 0; i < 3; i++ {
+		allHeaders, err := localNode.getNeighborsBlockHeaderByHeight(height, neighbors)
```
gdmmx, line 330: sleep(RetryDelay) was cancelled?

util/config/config.go
gdmmx: So, we should use a named const to replace it for code readability.

yilunzhang: This constant is defined for this purpose

node/syncblock.go
yilunzhang: Yes I don't think it's necessary anymore

gdmmx (approved):
LGTM

PR #741, opened 2020-09-22
Reorganize code structure to reduce go SDK memory usage

yilunzhang:


chain/blockvalidator.go
```diff
@@ -29,7 +32,12 @@ const (
 	NumGenesisBlocks           = 4
 )
 
-var timestampToleranceSalt []byte = util.RandomBytes(32)
+var (
+	ErrIDRegistered           = errors.New("ID has be registered")
```
bufrr, line 36: typo?

block/block.go
```diff
@@ -257,7 +257,7 @@ func (b *Block) GetInfo() ([]byte, error) {
 	info := &blockInfo{
 		Header:       unmarshaledHeader,
 		Transactions: make([]interface{}, 0),
-		Size:         b.ToMsgBlock().Size(),
+		Size:         proto.Size(b.ToMsgBlock()),
```
bufrr, line 260: why chosing `proto.Size` instead? And they both return the same value in this case right?

chain/blockvalidator.go
yilunzhang: You are right, fixed

block/block.go
yilunzhang: Google's pb only has this API, it returns the same result

bufrr (approved):

PR #746, opened 2020-10-07
Change subscribe meta internal type from string to bytes

yilunzhang:
Protobuf bytes and string have the same encoding but different restrictions: string can now only contain utf-8 characters. However, the current chain data contains non-utf-8 string in meta generated using old pb version, which will cause new version fail to save certain blocks. Changing internal type to bytes will resolve the issue without breaking compatibility.


bufrr:
What is the point of this modification?

yilunzhang:
The current version cannot save a few blocks containing non-uft8 subscribe metadata

pb/transaction.proto
```diff
@@ -72,7 +72,7 @@ message Subscribe {
 	string topic       = 3;
 	uint32 bucket      = 4 [deprecated = true];
 	uint32 duration    = 5;
-	string meta        = 6;
+	bytes meta        = 6;
```
bufrr, line 75: looks like unformatted

yilunzhang: Fixed

bufrr (approved):

PR #762, opened 2021-04-18
Init log module in nknc to prevent panic

yilunzhang:
This PR fix the bug that log module is not initialized and could cause panic if used.


gdmmx (approved):
LGTM

PR #787, opened 2021-05-10
Use voting topology for neighbor consensus state

yilunzhang:
This change will prevent some nodes from being stuck in wait for syncing state


gdmmx (approved):
LGTM

PR #789, opened 2021-05-11
Introduce sighash left shift mechanism for recent miners and skipped hop

yilunzhang:
This change will greatly enhance network security and reduce mining reward variance.

### DIFF
--- a/por/minercache.go
+++ b/por/minercache.go
@@ -52,6 +52,9 @@ func GetRecentMiner(blockHash []byte) (RecentMiner, error) {
 		if err != nil {
 			return nil, err
 		}
+		if header.UnsignedHeader.Height == 0 {
+			break
+		}
 		rm[string(header.UnsignedHeader.SignerPk)]++
 		hash = header.UnsignedHeader.PrevBlockHash
 	}
@@ -88,6 +91,9 @@ func GetSkipMiner(blockHash []byte) (SkipMiner, error) {
 		header, err := store.GetHeaderWithCache(hashUint256)
 		if err != nil {
 			return nil, err
+		}
+		if header.UnsignedHeader.Height == 0 {
+			break
 		}
 		winnerHash, err := common.Uint256ParseFromBytes(header.UnsignedHeader.WinnerHash)
 		if err != nil {
--- a/por/minercache.go
+++ b/por/minercache.go
@@ -1,22 +1,28 @@
 package por
 
 import (
+	"sort"
 	"sync"
 	"time"
 
 	"github.com/nknorg/nkn/v2/common"
 	"github.com/nknorg/nkn/v2/config"
+	"github.com/nknorg/nnet/overlay/chord"
 )
 
 var (
 	recentMinerLock  sync.Mutex
 	recentMinerCache common.Cache
+	skipMinerLock    sync.Mutex
+	skipMinerCache   common.Cache
 )
 
 type RecentMiner map[string]int
+type SkipMiner [][]byte
 
 func init() {
 	recentMinerCache = common.NewGoCache(5*time.Minute, time.Minute)
+	skipMinerCache = common.NewGoCache(5*time.Minute, time.Minute)
 }
 
 func GetRecentMiner(blockHash []byte) (RecentMiner, error) {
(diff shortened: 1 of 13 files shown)


por/minercache.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	recentMinerLock.Lock()
+	defer recentMinerLock.Unlock()
+
+	if v, ok := recentMinerCache.Get(blockHash); ok {
+		if rm, ok := v.(RecentMiner); ok {
+			return rm, nil
+		}
+	}
```
gdmmx, line 42: Why recentMinerCache.Get twice ?

yilunzhang: To prevent the cache from being calculated multiple times.

por/sigchain.go
```diff
@@ -185,3 +187,36 @@ func VerifySigChainPath(sc *pb.SigChain, height uint32) error {
...
+	}
+
+	if blockHash == common.EmptyUint256 {
+		return nil, errors.New("block hash in sigchain is empty")
+	}
+
+	height, err := store.GetHeightByBlockHash(blockHash)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
yilunzhang: This is needed if we don't want to rely on user to provide block height in `SignatureHashWithPenalty` function.

por/sigchain.go
```diff
@@ -185,3 +187,36 @@ func VerifySigChainPath(sc *pb.SigChain, height uint32) error {
...
+	leftShiftBit := 0
+	if config.SigChainRecentMinerBitShift.GetValueAtHeight(height) > 0 && config.SigChainRecentMinerBlocks > 0 {
+		rm, err := GetRecentMiner(sc.BlockHash)
+		if err != nil {
+			return nil, err
+		}
+
+		count := 0
+		for i := 0; i < sc.Length()-2; i++ {
+			count += rm[string(sc.Elems[i].NextPubkey)]
```
gdmmx, line 217: All keys of RecentMiner just only signerPk.
It means that there are only `config.SigChainRecentMinerBlocks` skipper check point. It is not enough maybe.

yilunzhang: That's the goal. We don't want add penalty for relayers that do not earn any reward.

bufrr: What if the key `string(sc.Elems[i].NextPubkey)` doesn't exist?

yilunzhang: The value will be 0

por/minercache.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		header, err := store.GetHeaderWithCache(hashUint256)
+		if err != nil {
+			return nil, err
+		}
+		rm[string(header.UnsignedHeader.SignerPk)]++
+		hash = header.UnsignedHeader.PrevBlockHash
+	}
+
+	recentMinerCache.Set(blockHash, rm)
```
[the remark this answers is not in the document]
yilunzhang: Use slice will have `O(N)` time complexity, while the current map has `O(1)` time complexity at the cost of a few MB more memory usage.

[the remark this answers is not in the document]
gdmmx: How about cache's value type is a &map[pk] pointer which refer to same RecentMiner[string] with 2*SigChainRecentMinerBlock keys? 
Just new another RecentMiner when block hash was forked.

* pseudo-code
 ```
if block fork {
    rm := make(RecentMiner, config.SigChainRecentMinerBlocks)
    for i := 0; i < config.SigChainRecentMinerBlocks; i++ {
        // refill RecentMiner
    }
    recentMinerCache.Set(blockHash, rm)

} else {
    existed_rm[currentHeader.SignerPk]++
    existed_rm.pop( 2*SigChainRecentMinerBlocks ahead header.SignerPk)

    recentMinerCache.Set(blockHash, existed_rm)
}
```

File: por/sigchain.go [hunk 5]
```diff
@@ -218,5 +220,37 @@ func SignatureHashWithPenalty(sc *pb.SigChain) ([]byte, error) {
...
+
+		leftShiftBit += count * int(config.SigChainSkipMinerBitShift.GetValueAtHeight(height))
+	}
+
 	return sc.SignatureHash(height, leftShiftBit)
 }
+
+func skipCount(from, to []byte, sm SkipMiner) int {
+	dist := chord.Distance(from, to, config.NodeIDBytes*8)
+	fingerIdx := dist.BitLen() - 1
```
gdmmx, line 245: Have we any distance check go ahead of skipCount?
e.g. bad case from == to from malicious sigchain.

por/sigchain.go
```diff
@@ -218,5 +220,37 @@ func SignatureHashWithPenalty(sc *pb.SigChain) ([]byte, error) {
...
+	fingerStartID := chord.PowerOffset(from, uint32(fingerIdx), config.NodeIDBytes*8)
+	beg := sort.Search(len(sm), func(i int) bool { return chord.CompareID(sm[i], fingerStartID) >= 0 })
+	end := sort.Search(len(sm), func(i int) bool { return chord.CompareID(sm[i], to) >= 0 })
+	skipped := 0
+	if chord.CompareID(fingerStartID, to) <= 0 {
+		skipped = end - beg
```
gdmmx, line 249: In `fingerStartID <= to` case, end would less than beg. It means that skipped is negative number.

yilunzhang: Nope, when fingerStartID <= to, beg <= end

gdmmx: range of beg = `Max(sm[:]...) - fingerStartID`
range of end = `Max(sm[:]...) - to`

`range of beg` should inclusive of `range of end` because fingerStartID is smaller.

yilunzhang: This is not correct. Please check what sort.Search does

gdmmx: My mistake. The `sort.Search` return index, not count.

PR #799, opened 2021-07-13
Update mainnet version url

yilunzhang:


bufrr (approved):

gdmmx (commented):
We should keep the older URL still available in order to compatible with those previous nknd.
```
curl -s https://nkn.org/mainnet.runtime.version  | wc
       0       0       0
```

yilunzhang:
> We should keep the older URL still available in order to compatible with those previous nknd.
> 
> ```
> curl -s https://nkn.org/mainnet.runtime.version  | wc
>        0       0       0
> ```

You forgot to add `-L` argument

PR #806, opened 2021-08-05
state fast sync

bufrr:


chain/blockvalidator.go
```diff
@@ -91,6 +91,18 @@ func TransactionCheck(ctx context.Context, block *block.Block) error {
...
 
+	txnsRoot, err := crypto.ComputeRoot(txnsHash)
+	if err != nil {
+		return fmt.Errorf("compute txns root error: %v", err)
+	}
+	if !bytes.Equal(txnsRoot.ToArray(), block.Header.UnsignedHeader.TransactionsRoot) {
+		return fmt.Errorf("computed txn root %x is different from txn root in header %x", txnsRoot.ToArray(), block.Header.UnsignedHeader.TransactionsRoot)
+	}
+
+	if DefaultLedger.Store.DuringFastSync(block.Header.UnsignedHeader.Height) {
```
yilunzhang, line 102: Txn hash is not affected by signature (same for block hash), so signature needs to be verified as well

chain/store/rollback.go
```diff
@@ -152,7 +152,7 @@ func (cs *ChainStore) rollbackSigChainCache(b *block.Block) error {
 }
 
 func (cs *ChainStore) rollbackDonation(b *block.Block) error {
-	if b.Header.UnsignedHeader.Height%uint32(config.RewardAdjustInterval) != 0 {
+	if b.Header.UnsignedHeader.Height%uint32(config.RewardAdjustInterval) != 0 || config.DonationNoDelay.GetValueAtHeight(b.Header.UnsignedHeader.Height) {
```
yilunzhang, line 155: You didn't rebase this commit

chain/store/store.go
```diff
@@ -52,6 +52,29 @@ func NewLedgerStore() (*ChainStore, error) {
...
+		log.Fatal("Trie traverse error: %v", err)
+	}
+
+	hs := sdb.trie.Hash()
+	if syncRootHash.CompareTo(hs) != 0 {
+		log.Fatal("State root not equal, should be: %v, we got: %v", syncRootHash.ToHexString(), hs.ToHexString())
+	}
+	cs.States = sdb
+
+	err = cs.persistSyncRootHeight(fastSyncHeight)
```
yilunzhang, line 72: Do you need to put a few other db items like `CurrentStateTrie`, `CurrentBlockHashKey`, etc? See logic in `persist()` function.

chain/store/store.go
```diff
@@ -354,27 +376,32 @@ func (cs *ChainStore) persist(b *block.Block) error {
...
 
-	err = cs.st.BatchPut(db.CurrentStateTrie(), root.ToArray())
-	if err != nil {
-		return err
+		err = cs.st.BatchPut(db.CurrentStateTrie(), root.ToArray())
+		if err != nil {
+			return err
+		}
+		defer func() {
+			cs.States = states
```
yilunzhang, line 398: Why move it to here? If db write failed (`Commit()` return error), `cs.states` should not be updated.

chain/store/store.go
```diff
@@ -354,27 +376,32 @@ func (cs *ChainStore) persist(b *block.Block) error {
...
+		}()
```
yilunzhang, line 399: Why the rest of this function should be run during fast sync?

chain/store/store.go
```diff
@@ -443,7 +468,7 @@ func (cs *ChainStore) SaveBlock(b *block.Block, pruning bool) error {
 		return err
 	}
 
-	if pruning {
+	if pruning && !cs.DuringFastSync(b.Header.UnsignedHeader.Height-1) {
```
yilunzhang, line 471: Why `-1` here?

chain/store/store.go
```diff
@@ -580,6 +605,10 @@ func (cs *ChainStore) GetDatabase() db.IStore {
 	return cs.st
 }
 
+func (cs *ChainStore) DuringFastSync(height uint32) bool {
+	return config.Parameters.SyncMode == "fast" && height <= cs.getSyncRootHeight()
```
yilunzhang, line 609: Genesis block height is 0, should handle it

chain/store/store.go
```diff
@@ -737,11 +775,15 @@ func (cs *ChainStore) GetPruningStartHeight() (uint32, uint32) {
...
 func (cs *ChainStore) getPruningStartHeight() (uint32, uint32) {
 	var pruningStartHeight, refCountStartHeight uint32
+	syncRootHeight := cs.getSyncRootHeight()
 
 	heightBuffer, err := cs.st.Get(db.TrieRefCountHeightKey())
 	if err != nil {
 		log.Info("get height of trie counted error:", err)
 		refCountStartHeight = 0
+		if syncRootHeight != 0 {
+			refCountStartHeight = syncRootHeight + 1
```
yilunzhang, line 785: It's better not modify logic here because it couples fast pruning logic with pruning logic. A better way is to set `TrieRefCountHeightKey`, `TrieCompactHeightKey`, etc after fast sync succeeds

chain/store/store.go
```diff
@@ -772,6 +818,25 @@ func (cs *ChainStore) persistCompactHeight(height uint32) error {
...
+func (cs *ChainStore) persistSyncRootHeight(height uint32) error {
+	heightBuffer := make([]byte, 4)
+	binary.LittleEndian.PutUint32(heightBuffer[:], height)
+	return cs.st.Put(db.TrieSyncRootHeightKey(), heightBuffer)
+}
+
+func (cs *ChainStore) getSyncRootHeight() uint32 {
+	heightBuffer, err := cs.st.Get(db.TrieSyncRootHeightKey())
+	if err != nil {
+		log.Info("get sync root height error:", err)
```
yilunzhang, line 830: Will this be printed when fast sync key is not set at all?

chain/trie/sync.go
```diff
@@ -0,0 +1,188 @@
...
+
+// request represents a scheduled or already in-flight state retrieval request.
+type request struct {
+	path []byte         // Merkle path leading to this node for prioritization
+	data []byte         // Data content of the node, cached until all subtrees complete
+	hash common.Uint256 // Hash of the node data content to retrieve
+}
+
+// schedule inserts a new state retrieval request into the fetch queue
+func (s *Sync) schedule(req *request) {
```
yilunzhang, line 22: Better to put method after Sync declaration. Let's follow this convention:

type definition

type constructor

type public method

type private method

chain/trie/sync.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	s.schedule(req)
+}
+
+// Pending returns the number of state entries currently pending for download.
+func (s *Sync) Pending() int {
+	return len(s.nodeReqs)
+}
+
+func (s *Sync) Missing(max int) (nodes []common.Uint256, paths []SyncPath) {
```
yilunzhang, line 70: Suggest change name to Pop and add comment

chain/trie/sync.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Sync is the main state trie synchronisation scheduler, which provides yet
+// unknown trie hashes to retrieve, accepts node data associated with said hashes
+// and reconstructs the trie step by step until all is done.
+type Sync struct {
+	membatch map[common.Uint256][]byte   // Memory buffer to avoid frequent database writes
+	nodeReqs map[common.Uint256]*request // Pending requests pertaining to a trie node hash
+	queue    *prque.Prque                // Priority queue with the pending requests
+	fetches  map[int]int                 // Number of active fetches per trie node depth
+	sync.Mutex
```
yilunzhang, line 40: What's this lock for? Seems not used anywhere

chain/trie/trie.go
```diff
@@ -334,7 +334,7 @@ func (t *Trie) Copy() *Trie {
 }
 
 func (t *Trie) TryTraverse() error {
-	return t.traverse(t.root, true)
+	return t.traverse(t.root, false)
```
yilunzhang, line 337: Why change this?

config/config.go
```diff
@@ -74,6 +74,9 @@ const (
 	MaxClientMessageSize            = 4 * 1024 * 1024
 	MinNameRegistrationFee          = 10 * common.StorageFactor
 	DefaultNameDuration             = 365 * 24 * 60 * 60 / int(ConsensusDuration/time.Second)
+	IdealBatchSize                  = 100 * 1024
```
yilunzhang, line 77: Suggest change name to `FastSyncMaxBatchSize`

config/config.go
```diff
@@ -74,6 +74,9 @@ const (
...
+	FastSyncPivot                   = 10
+	MaxStateReqSize                 = 500
```
yilunzhang, line 79: Suggest change name to `FastSyncMaxStateReqSize`

config/config.go
```diff
@@ -306,6 +314,7 @@ var (
 		SyncBlockHeaderRateBurst:     32768,
 		SyncBlockRateLimit:           256,
 		SyncBlockRateBurst:           1024,
+		SyncMode:                     "fast",
```
yilunzhang, line 317: Don't turn on fast sync by default now

node/downloader.go
```diff
@@ -0,0 +1,31 @@
+package node
+
+import (
+	"github.com/nknorg/nkn/v2/chain/db"
+	"github.com/nknorg/nkn/v2/util/log"
+
+	"github.com/nknorg/nkn/v2/common"
+)
+
+type Downloader struct {
```
yilunzhang, line 10: Do we even need this file and `Downloader`?

bufrr: For providing new peers and new pivot blockhash if syncer needed

chain/trie/trie.go
bufrr: Too many unuseful log when set `print` to true

yilunzhang: Then make `TryTraverse` accept a parameter because it's also used elsewhere

node/downloader.go
yilunzhang: Everything in this file can be compressed as
```
log.Info()
s := newStateSync(d, root, peers)
s.run()
```
in `StartFastSyncing()`

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+// it finishes, and finally notifying any goroutines waiting for the loop to
+// finish.
+func (s *stateSync) run() {
+	s.err = s.loop()
+	close(s.done)
+}
+
+// newStateSync creates a new state trie download scheduler. This method does not
+// yet start the sync. The user needs to call run to initiate.
+func newStateSync(d *Downloader, root common.Uint256, peers []*RemoteNode) *stateSync {
```
yilunzhang, line 171: Betrer move this to top, right below type definition

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+	hasher func([]byte) []byte
+
+	done chan struct{} // Channel to signal termination completion
+	err  error         // Any error hit during sync (set before completion)
+
+	hashesCh       chan []common.Uint256
+	stateCh        chan [][]byte // Response states channel
+	timeoutReq     chan []common.Uint256
+	failedHashesCh chan []common.Uint256
+	stopChans      []chan struct{}
```
yilunzhang, line 31: All workers can share the same stop channel, just close it when need to send stop signal

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+	root   common.Uint256 // State root currently being synced
+	sched  *trie.Sync     // State trie sync scheduler defining the tasks
+	hasher func([]byte) []byte
+
+	done chan struct{} // Channel to signal termination completion
+	err  error         // Any error hit during sync (set before completion)
+
+	hashesCh       chan []common.Uint256
+	stateCh        chan [][]byte // Response states channel
+	timeoutReq     chan []common.Uint256
```
yilunzhang, line 29: Seems not used

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+		if len(resp) > 0 {
+			s.stateCh <- resp
+		}
+	}
+}
+
+// run starts the task assignment and response processing loop, blocking until
+// it finishes, and finally notifying any goroutines waiting for the loop to
+// finish.
+func (s *stateSync) run() {
```
yilunzhang, line 164: We can merge `loop()` and `run()` into a `run()` that just return error

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+
+// loop is the main event loop of a state trie sync. It is responsible for the
+// assignment of new tasks to peers (including sending it to them) as well as
+// for the processing of inbound data.
+func (s *stateSync) loop() (err error) {
+	defer func() {
+		cerr := s.commit(true)
+		if err == nil {
+			err = cerr
+		}
```
yilunzhang, line 46: Should we stop here and return if cerr != nil?

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+		if err := s.d.database.BatchCommit(); err != nil {
+			err = cerr
+		}
+	}()
+
+	go s.startWorkerThread()
+
+	go func() {
+		for {
+			log.Info("pending states:", s.sched.Pending())
```
yilunzhang, line 60: Is this for debug only, or we plan to add this log for release version?

bufrr: For debug only

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+		}
+	}()
+
+	for s.sched.Pending() > 0 {
+		if err = s.commit(false); err != nil {
+			return err
+		}
+
+		for i := 0; i < len(s.peers); i++ {
```
yilunzhang, line 75: The current flow cannot make full use of resources. We should make this job dispatch part and the process result part parallel (in two different go routine). A few things also need to change:
* Use lock to make sure all public methods of trie.Sync are thread safe
* Use a channel to notify when new data is available in trie.Sync

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+	peersNum := len(s.peers)
+	for workerId = 0; workerId < uint32(peersNum); workerId++ {
+		wg.Add(1)
+		go func(workerId uint32) {
+			defer wg.Done()
+			time.Sleep(time.Duration(workerId) * syncWorkerStartInterval)
+			s.startWorker(ctx, workerId, s.peers[workerId%uint32(peersNum)])
+		}(workerId)
+	}
+	wg.Wait()
```
yilunzhang, line 200: Why block and wait?

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+	var workerId uint32
+	ctx := context.Background()
+
+	peersNum := len(s.peers)
+	for workerId = 0; workerId < uint32(peersNum); workerId++ {
+		wg.Add(1)
+		go func(workerId uint32) {
+			defer wg.Done()
+			time.Sleep(time.Duration(workerId) * syncWorkerStartInterval)
+			s.startWorker(ctx, workerId, s.peers[workerId%uint32(peersNum)])
```
yilunzhang, line 197: ctx seems not useful here at all

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+	var err error
+	res.Hash, err = common.Uint256ParseFromBytes(s.hasher(blob))
+	if err != nil {
+		return common.EmptyUint256, err
+	}
+	err = s.sched.Process(res)
+	return res.Hash, err
+}
+
+func (s *stateSync) startWorker(ctx context.Context, workerId uint32, p *RemoteNode) {
```
yilunzhang, line 129: Both ctx and workerId seems unnecessary

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+	var hashes []common.Uint256
+	for {
+		select {
+		case <-s.stopChans[workerId]:
+			return
+		case <-ctx.Done():
+			return
+		default:
```
yilunzhang, line 137: You should use `s.done` because it's the only channel that is closed when finish

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+func newStateSync(d *Downloader, root common.Uint256, peers []*RemoteNode) *stateSync {
+	return &stateSync{
+		d:              d,
+		root:           root,
+		sched:          trie.NewSync(root),
+		hasher:         trie.Sha256Key,
+		done:           make(chan struct{}),
+		stateCh:        make(chan [][]byte),
+		stopChans:      make([]chan struct{}, len(peers)),
+		hashesCh:       make(chan []common.Uint256, config.Parameters.SyncStatesBatchSize),
```
yilunzhang, line 180: This channel doesn't even need to have buffer

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+		failedHashesCh: make(chan []common.Uint256, config.Parameters.SyncStatesBatchSize),
```
yilunzhang, line 181: This channel is not needed at all. When fails, put it back to the priority queue

node/syncblock.go
```diff
@@ -334,10 +360,12 @@ func (localNode *LocalNode) getNeighborsMajorityBlockHashByHeight(height uint32,
 		}
 
 		counter := make(map[common.Uint256]int)
+		headers := make(map[common.Uint256]*block.Header)
 		totalCount := 0
 		allHeaders.Range(func(key, value interface{}) bool {
 			if header, ok := value.(*block.Header); ok && header != nil {
 				counter[header.Hash()]++
+				headers[header.Hash()] = value.(*block.Header)
```
yilunzhang, line 368: Need to check header signature before set this because hash does not depend on signature.

node/syncblock.go
```diff
@@ -326,6 +328,30 @@ func (localNode *LocalNode) getNeighborsBlockHeaderByHeight(height uint32, neigh
...
+// getNeighborsMajorityStateRootByHeight returns the majority of given
+// neighbors' block hash at a give height
+func (localNode *LocalNode) getNeighborsMajorityStateRootByHeight(height uint32, neighbors []*RemoteNode) common.Uint256 {
+	header := localNode.getNeighborsMajorityBlockByHeight(height, neighbors)
+	if header == nil {
+		return common.EmptyUint256
+	}
+	rootHash, err := common.Uint256ParseFromBytes(header.UnsignedHeader.StateRoot)
+	if err != nil {
+		log.Warningf("Parse uint256 from bytes at height %d error: %v", height, err)
```
yilunzhang, line 347: Should return `EmptyUint256`

config/config.go
```diff
@@ -74,6 +74,9 @@ const (
 	MaxClientMessageSize            = 4 * 1024 * 1024
 	MinNameRegistrationFee          = 10 * common.StorageFactor
 	DefaultNameDuration             = 365 * 24 * 60 * 60 / int(ConsensusDuration/time.Second)
+	IdealBatchSize                  = 100 * 1024
+	FastSyncPivot                   = 10
```
yilunzhang, line 78: Better remove this and use `RecentStateCount`

node/stateSync.go
```diff
@@ -0,0 +1,201 @@
...
+			return
+		case <-ctx.Done():
+			return
+		}
+
+		resp, err := p.GetStates(hashes)
+
+		if err != nil {
+			s.failedHashesCh <- hashes
+			continue
```
yilunzhang, line 153: We need to distinguish two scenarios: whether we should continue using this worker or not. A few examples:
* if it's timeout, we might give it a few retries before return
* in all other case, we should probably just return

node/syncblock.go
```diff
@@ -368,6 +400,21 @@ func (localNode *LocalNode) initSyncing() {
 // StartSyncing starts block syncing from current local ledger until it gets to
 // block height stopHeight with block hash stopHash from given neighbors
 func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHeight uint32, neighbors []*RemoteNode) (bool, error) {
+	cs := chain.DefaultLedger.Store.(*store.ChainStore)
+	fastSyncHeight := syncStopHeight - config.FastSyncPivot
+	log.Info("fast sync begin, stopHeight:", fastSyncHeight)
+	if config.Parameters.SyncMode == "fast" && cs.GetSyncRootHeight() == 0 {
```
yilunzhang, line 406: If a node restarts during fast sync, it cannot resume fast sync right?

bufrr: No

yilunzhang: We can use the following logic if `config.Parameters.SyncMode == "fast"`:
if `cs.GetSyncRootHeight() == 0`, then use computed `fastSyncHeight`; otherwise use `cs.GetSyncRootHeight()` as stop height. This way it can resume after restarts

bufrr: But it could be dangerous that if user restart node after long time, old root maybe got pruned already.

yilunzhang: In this case it's user's responsibility to remove chaindb

chain/blockvalidator.go
```diff
@@ -388,11 +392,6 @@ func HeaderCheck(b *block.Block) error {
 		return fmt.Errorf("signer check failed: %v", err)
 	}
 
-	err = SignatureCheck(header)
```
yilunzhang, line 391: This will skip signature check for ALL header check call and proposal signature is no longer checked. Instead of removing it, add a `IsSignatureChecked` field in header, just like `IsHeaderChecked` field in block.

chain/store/store.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/nknorg/nkn/v2/util/log"
 )
 
+const (
+	beforeFastSync uint32 = iota
```
yilunzhang, line 25: Don't use iota for persist state as it can change when you change code order. Use explicitly value

chain/store/store.go
```diff
@@ -52,6 +58,72 @@ func NewLedgerStore() (*ChainStore, error) {
 	return chain, nil
 }
 
+func (cs *ChainStore) PrepareFastSync(syncRootHash common.Uint256, fastSyncHeight uint32) {
```
yilunzhang, line 61: Let this function return err, and let the caller fatal

chain/store/store.go
```diff
@@ -52,6 +58,72 @@ func NewLedgerStore() (*ChainStore, error) {
...
+		log.Fatalf("Error persist state trie root: %v", err)
+	}
+
+	err = cs.st.BatchCommit()
+	if err != nil {
+		log.Fatalf("Error commit fast sync done: %v", err)
+	}
+}
+
+func (cs *ChainStore) FastSyncDone(syncRootHash common.Uint256) {
```
yilunzhang, line 98: Let this function return err, and let the caller fatal

chain/store/store.go
```diff
@@ -52,6 +58,72 @@ func NewLedgerStore() (*ChainStore, error) {
...
 }
 
+func (cs *ChainStore) PrepareFastSync(syncRootHash common.Uint256, fastSyncHeight uint32) {
+	err := cs.persistSyncRootHeight(fastSyncHeight)
+	if err != nil {
+		log.Fatalf("Error persist sync root height: %v", err)
+	}
+	refCountHeightBuffer := make([]byte, 4)
+	binary.LittleEndian.PutUint32(refCountHeightBuffer[:], fastSyncHeight+1)
+	err = cs.st.BatchPut(db.TrieRefCountHeightKey(), refCountHeightBuffer)
```
yilunzhang, line 68: Should set ref count height and pruned height AFTER fast sync is done.
Also I think it should be -1 instead of +1, same for pruned height

chain/store/store.go
```diff
@@ -52,6 +58,72 @@ func NewLedgerStore() (*ChainStore, error) {
...
+	}
+
+	fastSyncStatusBuf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(fastSyncStatusBuf[:], beginFastSync)
+	err = cs.st.BatchPut(db.TrieFastSyncStatusKey(), fastSyncStatusBuf)
+	if err != nil {
+		log.Fatalf("Error persist refCount height: %v", err)
+	}
+
+	err = cs.st.BatchPut(db.CurrentStateTrie(), syncRootHash.ToArray())
```
yilunzhang, line 87: Should be set AFTER fast sync finish

chain/blockvalidator.go
bufrr: I moved this check to the `SaveBlock` function

chain/store/store.go
bufrr: It would panic if we set this value after fast sync done when nknd restart, i need to test it to make sure

node/syncblock.go
```diff
@@ -368,6 +401,26 @@ func (localNode *LocalNode) initSyncing() {
 // StartSyncing starts block syncing from current local ledger until it gets to
 // block height stopHeight with block hash stopHash from given neighbors
 func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHeight uint32, neighbors []*RemoteNode) (bool, error) {
+	cs := chain.DefaultLedger.Store.(*store.ChainStore)
+	fastSyncHeight := syncStopHeight - config.Parameters.RecentStateCount
+	log.Info("fast sync begin, stopHeight:", fastSyncHeight)
```
yilunzhang, line 406: Move this log inside if

node/syncblock.go
```diff
@@ -368,6 +401,26 @@ func (localNode *LocalNode) initSyncing() {
 // StartSyncing starts block syncing from current local ledger until it gets to
 // block height stopHeight with block hash stopHash from given neighbors
 func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHeight uint32, neighbors []*RemoteNode) (bool, error) {
+	cs := chain.DefaultLedger.Store.(*store.ChainStore)
+	fastSyncHeight := syncStopHeight - config.Parameters.RecentStateCount
```
yilunzhang, line 405: Should use fast sync height stored in db is stored

chain/store/store.go
```diff
@@ -779,6 +856,47 @@ func (cs *ChainStore) persistCompactHeight(height uint32) error {
...
+		return beforeFastSync
+	}
+	return binary.LittleEndian.Uint32(statusBuffer)
+}
+
+func (cs *ChainStore) GetFastSyncStatus() uint32 {
+	return cs.getFastSyncStatus()
+}
+
+func (cs *ChainStore) FastSyncRestorable() bool {
```
yilunzhang, line 896: Change this function to `ShouldFastSync` and change logic to:
if state is beforeFastSync, return true only if local block height == 0
if state is beginFastSync, return true
if state is postFastSync, return false

node/syncblock.go
```diff
@@ -368,6 +401,26 @@ func (localNode *LocalNode) initSyncing() {
...
+	log.Info("fast sync begin, stopHeight:", fastSyncHeight)
+	if config.Parameters.SyncMode == "fast" && cs.FastSyncRestorable() {
+		fastSyncRootHash, err := cs.GetCurrentBlockStateRoot()
+		if err != nil {
+			log.Infof("no state root found before fast sync, will processing a fresh fast sync: %v", err)
```
yilunzhang, line 410: No need to print err if it's normal

node/syncblock.go
```diff
@@ -368,6 +401,26 @@ func (localNode *LocalNode) initSyncing() {
 // StartSyncing starts block syncing from current local ledger until it gets to
 // block height stopHeight with block hash stopHash from given neighbors
 func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHeight uint32, neighbors []*RemoteNode) (bool, error) {
+	cs := chain.DefaultLedger.Store.(*store.ChainStore)
+	fastSyncHeight := syncStopHeight - config.Parameters.RecentStateCount
+	log.Info("fast sync begin, stopHeight:", fastSyncHeight)
+	if config.Parameters.SyncMode == "fast" && cs.FastSyncRestorable() {
+		fastSyncRootHash, err := cs.GetCurrentBlockStateRoot()
```
yilunzhang, line 408: Shouldn't use current state root. Consider this case: node stop during fast sync, then disable fast sync and restart. Change current state root will cause node to have wrong state without fast sync enabled.

chain/trie/sync.go
```diff
@@ -0,0 +1,213 @@
...
+	defer s.Unlock()
+	req := &request{
+		path: path,
+		hash: root,
+	}
+	s.schedule(req)
+}
+
+// Pending returns the number of state entries currently pending for download.
+func (s *Sync) Pending() int {
```
yilunzhang, line 63: RLock

chain/trie/sync.go
```diff
@@ -0,0 +1,213 @@
...
+// Pop return the specific number of hash jobs from heap
+func (s *Sync) Pop(max int) ([]common.Uint256, []int64) {
+	s.Lock()
+	defer s.Unlock()
+	var (
+		nodeHashes []common.Uint256
+		prios      []int64
+	)
+
+	for !s.queue.Empty() && (max == 0 || len(nodeHashes) < max) {
```
yilunzhang, line 76: Use `empty()`

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+
+	for s.sched.Pending() > 0 {
+		if err := s.commit(false); err != nil {
+			return err
+		}
+		select {
+		case data := <-s.stateCh:
+			for _, blob := range data {
+				_, err := s.processNodeData(blob)
+				switch err {
```
yilunzhang, line 87: No need to use switch for this....

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+				case nil:
+					s.bytesUncommitted += len(blob)
+				default:
+					log.Info("processNodeData err:", err)
```
yilunzhang, line 91: Should we do something when error happens?

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+	var hashes []common.Uint256
+	fail := 0
+	for {
+		select {
+		case <-s.done:
+			return
+		default:
+		}
+
+		select {
```
yilunzhang, line 144: You can combine this two select

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+		case hashes = <-s.hashesCh:
+		case <-s.done:
+			return
+		}
+
+		resp, err := p.GetStates(hashes)
+
+		if err != nil {
+			s.sched.Push(hashes, int64(^uint(0)>>1)) // max int64 impl before go 1.17
```
yilunzhang, line 153: Use `math.Maxint64`

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+			select {}
```
yilunzhang, line 154: What's this?

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+			fail++
+			if fail >= 3 {
```
yilunzhang, line 156: Use constant `maxSyncWorkerFails`

node/stateSync.go
```diff
@@ -0,0 +1,184 @@
...
+func (s *stateSync) run() error {
+	err := s.loop()
+	close(s.done)
+	return err
+}
+
+func (s *stateSync) startWorkerThread() {
+	var workerId uint32
+	peersNum := len(s.peers)
+	for workerId = 0; workerId < uint32(peersNum); workerId++ {
```
yilunzhang, line 179: Can change to `concurrentSyncRequestPerNeighbor * peersNum`

chain/db/prefix.go
```diff
@@ -30,6 +30,9 @@ const (
 	TRIE_RefCountHeight DataEntryPrefix = 0xa2
 	TRIE_PrunedHeight   DataEntryPrefix = 0xa3
 	TRIE_CompactHeight  DataEntryPrefix = 0xa4
+	TRIE_SyncRootHeight DataEntryPrefix = 0xa5
```
yilunzhang, line 33: Suggest change to `TRIE_FastSyncRootHeight`

chain/db/prefix.go
```diff
@@ -92,3 +99,11 @@ func TriePrunedHeightKey() []byte {
 func TrieCompactHeightKey() []byte {
 	return paddingKey(TRIE_CompactHeight, nil)
 }
+
+func TrieSyncRootHeightKey() []byte {
```
yilunzhang, line 103: Suggest rename to `TrieFastSyncRootHeightKey`

chain/store/store.go
```diff
@@ -576,10 +656,35 @@ func (cs *ChainStore) GetCurrentBlockStateRoot() (common.Uint256, error) {
...
+	}
+
+	return hash, nil
+}
+
 func (cs *ChainStore) GetDatabase() db.IStore {
 	return cs.st
 }
 
+func (cs *ChainStore) DuringFastSync(height uint32) bool {
```
yilunzhang, line 677: Suggest rename to `IsDuringFastSync` and only use this function once at the block sync logic, and pass it along the call just like `pruning`

chain/trie/pruning.go
```diff
@@ -16,14 +16,16 @@ type RefCounts struct {
 
 	targetRefCountHeight uint32
 	targetPruningHeight  uint32
+	syncRootHeight       uint32
 }
 
-func NewRefCounts(t *Trie, refCountTargetHeight, pruningTargetHeight uint32) (*RefCounts, error) {
+func NewRefCounts(t *Trie, refCountTargetHeight, pruningTargetHeight, syncRootHeight uint32) (*RefCounts, error) {
```
yilunzhang, line 22: Why pass syncRootHeight?

chain/store/store.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/nknorg/nkn/v2/util/log"
 )
 
+const (
+	beforeFastSync uint32 = 0
```
yilunzhang, line 25: maybe we can just use byte instead of uint32

chain/store/store.go
```diff
@@ -52,6 +58,81 @@ func NewLedgerStore() (*ChainStore, error) {
 	return chain, nil
 }
 
+func (cs *ChainStore) PrepareFastSync(fastSyncHeight uint32) error {
+	err := cs.persistSyncRootHeight(fastSyncHeight)
+	if err != nil {
+		return err
+	}
+
+	fastSyncStatusBuf := make([]byte, 4)
```
yilunzhang, line 67: You have `persistSyncStatus` method

chain/store/store.go
```diff
@@ -52,6 +58,81 @@ func NewLedgerStore() (*ChainStore, error) {
...
+		return err
+	}
+
+	hs := sdb.trie.Hash()
+	if syncRootHash.CompareTo(hs) != 0 {
+		log.Fatal("State root not equal, should be: %v, we got: %v", syncRootHash.ToHexString(), hs.ToHexString())
+	}
+	cs.States = sdb
+
+	refCountHeightBuffer := make([]byte, 4)
```
yilunzhang, line 98: Better to create a refcount and use the existing `PersistRefCountHeights()` method

chain/store/store.go
```diff
@@ -52,6 +58,81 @@ func NewLedgerStore() (*ChainStore, error) {
...
+	binary.LittleEndian.PutUint32(refCountHeightBuffer[:], fastSyncHeight-1)
+	err = cs.st.BatchPut(db.TrieRefCountHeightKey(), refCountHeightBuffer)
+	if err != nil {
+		return err
+	}
+
+	pruneHeightBuffer := make([]byte, 4)
```
yilunzhang, line 105: Better to create a refcount and use the existing `PersistPrunedHeights()` method

chain/store/store.go
```diff
@@ -52,6 +58,81 @@ func NewLedgerStore() (*ChainStore, error) {
...
+	binary.LittleEndian.PutUint32(pruneHeightBuffer[:], fastSyncHeight-1)
+	err = cs.st.BatchPut(db.TriePrunedHeightKey(), pruneHeightBuffer)
+	if err != nil {
+		return err
+	}
+
+	fastSyncStatusBuf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(fastSyncStatusBuf[:], postFastSync)
```
yilunzhang, line 113: Use persistSyncStatus method

chain/store/store.go
```diff
@@ -354,40 +434,42 @@ func (cs *ChainStore) persist(b *block.Block) error {
...
+			err = donation.Serialize(w)
+			if err != nil {
+				return err
+			}
+
+			if err := cs.st.BatchPut(db.DonationKey(b.Header.UnsignedHeader.Height), w.Bytes()); err != nil {
+				return err
+			}
+		}
+		cs.States = states
```
yilunzhang, line 472: Set states at the end just like before

chain/store/store.go
```diff
@@ -576,10 +656,35 @@ func (cs *ChainStore) GetCurrentBlockStateRoot() (common.Uint256, error) {
...
+
+	return hash, nil
+}
+
 func (cs *ChainStore) GetDatabase() db.IStore {
 	return cs.st
 }
 
+func (cs *ChainStore) DuringFastSync(height uint32) bool {
+	if height == 0 {
```
yilunzhang, line 678: Don't need this if `getSyncRootHeight()` can return error

chain/store/store.go
```diff
@@ -779,6 +886,53 @@ func (cs *ChainStore) persistCompactHeight(height uint32) error {
...
+	if err != nil {
+		log.Info("no previous fast sync status found")
+		return beforeFastSync
+	}
+	return binary.LittleEndian.Uint32(statusBuffer)
+}
+
+func (cs *ChainStore) ShouldFastSync() bool {
+	status := cs.getFastSyncStatus()
+	if status == beforeFastSync && cs.currentBlockHeight == 0 {
```
yilunzhang, line 924: Better to use GetHeight()

yilunzhang: This is also a good place to use switch

node/syncblock.go
```diff
@@ -368,7 +406,44 @@ func (localNode *LocalNode) initSyncing() {
...
 func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHeight uint32, neighbors []*RemoteNode) (bool, error) {
-	var err error
+	cs := chain.DefaultLedger.Store.(*store.ChainStore)
+
+	fastSyncHeight, err := cs.GetSyncRootHeight()
+	if err != nil {
+		log.Info("local sync root height not found:", err)
+	}
+	if config.Parameters.SyncMode != "fast" && fastSyncHeight > cs.GetHeight() {
+		log.Fatalf("invalid sync mode %s at height %d", config.Parameters.SyncMode, cs.GetHeight())
```
yilunzhang, line 416: This will cause problem if user stop node during fast sync, turn off fast sync and restart node

bufrr: User can not stop fast sync and restart as full sync, because old stat didn't not exist which needed by full sync

node/syncblock.go
```diff
@@ -368,7 +406,44 @@ func (localNode *LocalNode) initSyncing() {
 // StartSyncing starts block syncing from current local ledger until it gets to
 // block height stopHeight with block hash stopHash from given neighbors
 func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHeight uint32, neighbors []*RemoteNode) (bool, error) {
-	var err error
```
yilunzhang, line 371: Put everything inside syncOnce.Do just like previous syncing logic

node/syncblock.go
```diff
@@ -368,7 +406,44 @@ func (localNode *LocalNode) initSyncing() {
...
+
+	if config.Parameters.SyncMode == "fast" && cs.ShouldFastSync() {
+		if fastSyncHeight == 0 {
+			fastSyncHeight = syncStopHeight - config.Parameters.RecentStateCount
+		}
+		fastSyncRootHash, err := cs.GetFastSyncStateRoot()
+		if err != nil {
+			fastSyncRootHash = localNode.GetNeighborsMajorityStateRootByHeight(fastSyncHeight, neighbors)
+			if fastSyncRootHash == common.EmptyUint256 {
+				log.Fatalf("get neighbors majority state root of height %d failed", fastSyncHeight)
```
yilunzhang, line 427: No need to fatal, when put inside syncOnce.Do, just set err and return, the caller will falta

node/syncblock.go
```diff
@@ -334,10 +360,18 @@ func (localNode *LocalNode) getNeighborsMajorityBlockHashByHeight(height uint32,
...
+		headers := make(map[common.Uint256]*block.Header)
 		totalCount := 0
 		allHeaders.Range(func(key, value interface{}) bool {
 			if header, ok := value.(*block.Header); ok && header != nil {
 				counter[header.Hash()]++
+				if _, ok := headers[header.Hash()]; !ok {
+					h := value.(*block.Header)
+					err = h.VerifySignature()
+					if err == nil {
+						headers[header.Hash()] = h
```
yilunzhang, line 372: We can add a log here when err != nil, like
```
log.Infof("Received header with invalid signature from neighbor %s", key)
```

node/syncblock.go
```diff
@@ -376,6 +414,48 @@ func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHe
...
 
 	syncOnce.Do(func() {
+		cs := chain.DefaultLedger.Store.(*store.ChainStore)
+
+		fastSyncHeight, err := cs.GetSyncRootHeight()
+		if err != nil {
+			log.Info("local sync root height not found:", err)
+		}
+		if config.Parameters.SyncMode != "fast" && fastSyncHeight > cs.GetHeight() {
+			err = fmt.Errorf("invalid sync mode %s at height %d", config.Parameters.SyncMode, cs.GetHeight())
```
yilunzhang, line 424: Need to return

node/syncblock.go
```diff
@@ -376,6 +414,48 @@ func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHe
...
+		}
+
+		if config.Parameters.SyncMode == "fast" && cs.ShouldFastSync() {
+			if fastSyncHeight == 0 {
+				fastSyncHeight = syncStopHeight - config.Parameters.RecentStateCount
```
yilunzhang, line 429: Need to check if `syncStopHeight > config.Parameters.RecentStateCount`. If not, we shouldn't use fast sync at all

node/syncblock.go
```diff
@@ -376,6 +414,48 @@ func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHe
...
+					return
+				}
+			}
+
+			err = cs.PrepareFastSync(fastSyncHeight, fastSyncRootHash)
+			if err != nil {
+				err = fmt.Errorf("prepare fast sync err: %v", err)
+				return
+			}
+			localNode.SetSyncState(pb.SyncState_SYNC_STARTED)
```
yilunzhang, line 445: Don't set it here. Move all fast sync logic to L461, after sync state is set

node/syncblock.go
```diff
@@ -376,6 +414,48 @@ func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHe
 	localNode.mu.RUnlock()
 
 	syncOnce.Do(func() {
+		cs := chain.DefaultLedger.Store.(*store.ChainStore)
```
yilunzhang, line 417: Don't do manual type conversion. If needed, add method to Store interface

node/syncblock.go
```diff
@@ -392,7 +472,7 @@ func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHe
 
 		_, err = localNode.maybeRollback(neighbors)
 		if err != nil {
-			log.Fatalf("Rollback error: %v", err)
+			err = fmt.Errorf("rollback error: %v", err)
```
yilunzhang, line 475: need return

node/syncblock.go
```diff
@@ -378,6 +416,50 @@ func (localNode *LocalNode) StartSyncing(syncStopHash common.Uint256, syncStopHe
...
+		if config.Parameters.SyncMode != "fast" && fastSyncHeight > cs.GetHeight() {
+			err = fmt.Errorf("invalid sync mode %s at height %d", config.Parameters.SyncMode, cs.GetHeight())
+			return
+		}
+
+		if config.Parameters.SyncMode == "fast" && cs.ShouldFastSync() {
+			if fastSyncHeight == 0 {
+				if syncStopHeight < config.Parameters.RecentStateCount {
+					err = fmt.Errorf("not enough height to use fast sync %d/%d", syncStopHeight, config.Parameters.RecentStateCount)
+					return
```
yilunzhang, line 433: In this case node should use normal sync, not return error

chain/store/store.go
```diff
@@ -911,7 +911,10 @@ func (cs *ChainStore) getFastSyncStatus() byte {
 	return status[0]
 }
 
-func (cs *ChainStore) ShouldFastSync() bool {
+func (cs *ChainStore) ShouldFastSync(syncStopHeight uint32) bool {
+	if syncStopHeight <= config.Parameters.RecentStateCount {
```
yilunzhang, line 915: Only if status is beforeFastSync

yilunzhang:
For those temporary hard coded constant for testing, better to remove them from commit history instead of pushing a new commit

PR #813, opened 2021-08-18
Node will get all neighbors state instead of just voting neighbors

yilunzhang:

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -100,14 +100,21 @@ func (consensus *Consensus) getNeighborConsensusState(neighbor *node.RemoteNode)
 	return replyMsg, nil
 }
 
-// getAllNeighborsConsensusState returns the latest block info of all neighbors
-// by calling getNeighborConsensusState on all of them concurrently.
-func (consensus *Consensus) getAllNeighborsConsensusState() (map[string]*pb.GetConsensusStateReply, error) {
-	allNeighbors := consensus.localNode.GetVotingNeighbors(nil)
-	if len(allNeighbors) < minConsensusStateNeighbors {
-		allNeighbors = consensus.localNode.GetNeighbors(nil)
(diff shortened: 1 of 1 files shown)


consensus/state.go
```diff
@@ -100,14 +100,21 @@ func (consensus *Consensus) getNeighborConsensusState(neighbor *node.RemoteNode)
...
+func (consensus *Consensus) getVotingNeighborsConsensusState() (map[string]*pb.GetConsensusStateReply, error) {
+	allNeighbors := consensus.localNode.GetNeighbors(nil)
+	votingNeighbors := consensus.localNode.GetVotingNeighbors(nil)
+	if len(votingNeighbors) < minConsensusStateNeighbors {
+		votingNeighbors = allNeighbors
 	}
-	allStates := make(map[string]*pb.GetConsensusStateReply, len(allNeighbors))
+	votingNeighborsMap := make(map[string]struct{}, len(votingNeighbors))
+	for _, neighbor := range votingNeighbors {
+		votingNeighborsMap[neighbor.GetID()] = struct{}{}
```
bufrr, line 114: Why we need to initialize `votingNeighborsMap` , we may have key with empty value, which previous code wont have

yilunzhang: This is map not slice, make will not create empty value

consensus/state.go
```diff
@@ -100,14 +100,21 @@ func (consensus *Consensus) getNeighborConsensusState(neighbor *node.RemoteNode)
...
+	}
+
+	states := make(map[string]*pb.GetConsensusStateReply, len(allNeighbors))
```
bufrr, line 117: maybe same length as `votingNeighbors`

yilunzhang: Fixed

bufrr (approved):

PR #833, opened 2022-01-12
Add connTime to neighbor/chord rpc response 

yilunzhang:
This PR add connTime field to get neighbors and get chord ring rpc response

### DIFF
--- a/node/info.go
+++ b/node/info.go
@@ -12,7 +12,17 @@ import (
 	"github.com/nknorg/nnet/overlay/chord"
 )
 
-type ChordRemoteNodeInfo nnetnode.RemoteNode
+type ChordRemoteNodeInfo struct {
+	localNode  *LocalNode
+	remoteNode *nnetnode.RemoteNode
+}
+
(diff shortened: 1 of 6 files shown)


node/info.go
```diff
@@ -39,17 +49,24 @@ func (rn *ChordRemoteNodeInfo) MarshalJSON() ([]byte, error) {
...
+	out["id"] = hex.EncodeToString(crn.remoteNode.Node.Node.Id)
+	out["isOutbound"] = crn.remoteNode.IsOutbound
+	out["roundTripTime"] = crn.remoteNode.GetRoundTripTime() / time.Millisecond
 	out["protocolVersion"] = nodeData.ProtocolVersion
 
+	nbr := crn.localNode.getNeighborByNNetNode(crn.remoteNode)
+	if nbr != nil {
+		out["connTime"] = time.Since(nbr.Node.startTime) / time.Second
+	} else {
+		out["connTime"] = 0
```
bufrr, line 67: Does it make sense when `nbr == nil` and we still have node info like `roundTripTime`?

bufrr: In this case I prefer to mv L67 to L62

bufrr (approved):

node/info.go
yilunzhang: Done

PR #837, opened 2022-02-17
Compute relay priority based on connTime

yilunzhang:
This is the initial implementation of NKP-0021

### DIFF
--- a/node/localnode.go
+++ b/node/localnode.go
@@ -166,6 +166,16 @@ func NewLocalNode(wallet *vault.Wallet, nn *nnet.NNet) (*LocalNode, error) {
 
 	nn.MustApplyMiddleware(routing.RemoteMessageRouted{localNode.remoteMessageRouted, 0})
 
+	nn.MustApplyMiddleware(chord.RelayPriority{func(rn *nnetnode.RemoteNode, priority float64) (float64, bool) {
+		var connTime time.Duration
+		nbr := localNode.getNeighborByNNetNode(rn)
+		if nbr != nil {
+			connTime = time.Since(nbr.Node.startTime)
+		}
+		priority = relayPriority(priority, connTime)
+		return priority, true
+	}, 0})
+
 	return localNode, nil
 }
 
--- a/cmd/nknd/nknd.go
+++ b/cmd/nknd/nknd.go
@@ -47,7 +47,7 @@ import (
 )
 
 const (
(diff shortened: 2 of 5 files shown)


node/localnode.go
```diff
@@ -166,6 +166,16 @@ func NewLocalNode(wallet *vault.Wallet, nn *nnet.NNet) (*LocalNode, error) {
...
 	nn.MustApplyMiddleware(routing.RemoteMessageRouted{localNode.remoteMessageRouted, 0})
 
+	nn.MustApplyMiddleware(chord.RelayPriority{func(rn *nnetnode.RemoteNode, priority float64) (float64, bool) {
+		var connTime time.Duration
+		nbr := localNode.getNeighborByNNetNode(rn)
+		if nbr != nil {
+			connTime = time.Since(nbr.Node.startTime)
+		}
+		priority = relayPriority(priority, connTime)
+		return priority, true
```
bufrr, line 176: This middleware will always return `true` and no `remotenode` will be discarded in nnet lib

bufrr (approved):

rdavis20:
> > Is uptime a consensus? Can it be circumvented by modifying the code?
> > like:
> > time.Now - 72 * 3600?
> 
> Modify a node's code only affect how it chooses next hop. It won't affect how other nodes choose it.

you're wrong, I know more about nkn than you

lhllhx:
> > Is uptime a consensus? Can it be circumvented by modifying the code?
> > like:
> > time.Now - 72 * 3600?
> 
> Each node only relies on the uptime measured by himself, not the time from its neighbors. Modify a node's code only affect how it chooses next hop. It won't affect how other nodes choose it.

If someone have large number nodes, and he use the modify version. Will he get a higher probability of getting a block?

hiDandelion:
> > > Is uptime a consensus? Can it be circumvented by modifying the code?
> > > like:
> > > time.Now - 72 * 3600?
> > 
> > 
> > Each node only relies on the uptime measured by himself, not the time from its neighbors. Modify a node's code only affect how it chooses next hop. It won't affect how other nodes choose it.
> 
> If someone have large number nodes, and he use the modify version. Will he get a higher probability of getting a block?

不可能，绝对不可能。

yilunzhang:
> If someone have large number nodes, and he use the modify version. Will he get a higher probability of getting a block?

The impact should be neglectable as the chance of each connection to be affected is only O(p^2) where p is his node count over total network size

PR #845, opened 2022-03-28
Change txn sorting to fee per size

yilunzhang:


transaction/sort.go
```diff
@@ -0,0 +1,78 @@
...
+func (s *SortTxnsByFee) Pop() interface{} {
+	old := *s
+	n := len(old)
+	x := old[n-1]
+	*s = old[0 : n-1]
+	return x
+}
+
+func CompareTxnsByFeePerSize(txn1, txn2 *Transaction) int {
+	feePerSize1 := float64(txn1.UnsignedTx.Fee) / (float64(txn1.GetSize()) + 0.01)
```
bufrr, line 55: What does this `0.01` do?

yilunzhang: Prevent panic when txn size is zero, it shouldn't happen, just prevent malicious transaction

bufrr: > 

But malicious transaction with zero size will have higher rank with very small tx fee

yilunzhang: No valid transaction will have be smaller than generate ID transaction v1. This is just to prevent invalid transaction from making node panic

bufrr (approved):

gdmmx (approved):
LGTM

transaction/sort.go
```diff
@@ -37,6 +43,10 @@ func CompareTxnsByFee(txn1, txn2 *Transaction) int {
 	return 0
 }
 
+func IsLowFeeTxn(txn *Transaction) bool {
+	return txn.UnsignedTx.Fee < config.Parameters.LowTxnFee
+}
+
```
gdmmx, line 49: `IsLowFeeTxn` was defined but no one used it.

yilunzhang: It's reversed in case we need to change back in the future

gdmmx (approved):
LGTM

PR #857, opened 2022-05-15
Rewrite CLI: nknc and nknd

omani:
Switch from urfave/cli to spf13/cobra. Same functionality, just different cli library, resulting in less loc.

Additionally, minor changes in code to fix a few warnings by compiler.

### Extra information
cobra will have a new release around June 2022 with new features like `MarkFlagsRequiredTogether()` and `MarkFlagsMutuallyExclusive()` which will reduce the loc even further, since we won't have to check for set flags manually anymore.

Im planning to implement them once cobra releases v1.5.0.


cmd/nknd/commands/root.go
```diff
@@ -43,94 +41,66 @@ import (
...
-		if err == nil {
-			return addr, err
-		}
-		log.Warningf("Ask my ID from %s error: %v", seed, err)
-	}
-	return "", errors.New("Tried all seeds but can't got my external IP and nknID")
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
+	rootCmd.Flags().BoolVarP(&createMode, "create", "c", false, "Create Mode")
+	rootCmd.Flags().StringVar(&config.SeedList, "seed", "", "Seed node address to join, multiple seeds should be split by comma")
+	rootCmd.Flags().StringVarP(&password.Passwd, "passwd", "p", "", "Password of Your wallet private Key")
```
gdmmx, line 85: Deprecated -p option for password since it should be showed in `ps` or others process info.

cmd/nknd/commands/root.go
```diff
@@ -375,139 +422,18 @@ func GetRemoteVersionNum() (int, error) {
...
-			if verNum > NetVersionNum {
-				log.Fatal("Your current nknd is deprecated, Please download the latest NKN software from https://github.com/nknorg/nkn/releases")
-			}
-
+	for range timer.C {
+		verNum, err := GetRemoteVersionNum()
+		if err != nil {
+			log.Warningf("Get the remote version number error: %v", err)
 			timer.Reset(30 * time.Minute)
+			break
```
yilunzhang, line 430: `break` should be changed to `continue` since you removed `select`

yilunzhang:
I think it's better to submit the changes for nknc in the same PR so we can remove the dependency of `urfave/cli`.

omani:
pushed nknc cli. fixed review.

cmd/nknd/commands/root.go
```diff
@@ -43,6 +43,7 @@ import (
 	"github.com/nknorg/nnet/overlay/chord"
 	"github.com/rdegges/go-ipify"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
```
gdmmx, line 46: `imported and not used: "github.com/spf13/pflag"` in my work tree.

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,216 @@
...
+	precision uint32
+)
+
+func init() {
+	rootCmd.AddCommand(assetCmd)
+
+	assetCmd.Flags().BoolVarP(&issue, "issue", "i", false, "issue asset")
+	assetCmd.Flags().BoolVarP(&transfer, "transfer", "t", false, "transfer asset")
+	assetCmd.Flags().StringVarP(&WalletFile, "wallet", "w", "", "wallet name")
+	assetCmd.Flags().StringVar(&WalletPassword, "password", "", "wallet password")
```
gdmmx, line 60: There is default value for `WalletFile` from `config.Parameters.WalletFile`.
`assetCmd` used global var will lost  the default value.

Not only `WalletFile` flag in `assetCmd`, but also several flags in all `sub commands` lost their default value.

cmd/nknd/commands/root.go
omani: that's right. strange my vscode didnt notify me here. thanks.

omani: yes. I missed that one.

omani:
squashed.

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,216 @@
...
+	assetCmd.Flags().StringVarP(&value, "value", "v", "", "asset amount in transfer asset or totalSupply in inssue assset")
+	assetCmd.Flags().StringVarP(&fee, "fee", "f", "", "transaction fee")
+	assetCmd.Flags().Uint64Var(&Nonce, "nonce", 0, "nonce")
+	assetCmd.Flags().StringVar(&assetname, "name", "", "asset name")
+	assetCmd.Flags().StringVar(&symbol, "symbol", "", "asset symbol")
+	assetCmd.Flags().Uint32Var(&precision, "precision", 0, "asset precision")
+
+	assetCmd.Flags().MarkHidden("password")
+	assetCmd.MarkFlagRequired("wallet")
+	assetCmd.MarkFlagRequired("password")
```
gdmmx, line 71: I think these two flags should not mark required since we allow from `default value in config` and from `gopass.GetPasswd`.

omani: fixed.

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/nknorg/nkn/v2/config"
+	"github.com/nknorg/nkn/v2/util/log"
+	"github.com/nknorg/nkn/v2/vault"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+const (
+	RANDBYTELEN = 4
+)
```
gdmmx, line 21: It seems that this global var wasn't used at anywhere?

omani: yes. it was there before so I didnt remove it. this is only a change of the underlying cli lib. no logic in code has been touched. we gotta ask @yilunzhang

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func getPassword(passwd string) []byte {
+	var tmp []byte
+	var err error
+
+	if passwd != "" {
+		tmp = []byte(passwd)
+	} else {
+		fmt.Printf("%s:", "Password")
```
bufrr, line 206: Should not print password

omani: take a look again. it prints a string "Password".

cmd/nknc/commands/debug.go
```diff
@@ -0,0 +1,54 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		showusage := true
+		cmd.Flags().Visit(func(name *pflag.Flag) {
+			showusage = false
+		})
+		if showusage {
+			return cmd.Usage()
+		}
+		if err := debugAction(); err != nil {
+			log.Error(err)
```
bufrr, line 27: Why not return this err?

omani: because cobra would show the usage help text if I return err.

bufrr: > because cobra would show the usage help text if I return err.

Then this function will never return any err even it's return type is error?

cmd/nknc/commands/asset.go
yilunzhang: No worries, you can leave it there and we will clean it up later.

cmd/nknc/commands/debug.go
omani: `cmd.Usage()` returns an error.

cmd/nknc/commands/id.go
```diff
@@ -0,0 +1,137 @@
...
+	idCmd.Flags().StringVarP(&WalletFile, "wallet", "w", config.Parameters.WalletFile, "wallet name")
+	idCmd.Flags().StringVarP(&WalletPassword, "password", "p", "", "wallet password")
+	idCmd.Flags().StringVar(&regfee, "regfee", "", "registration fee")
+	idCmd.Flags().StringVarP(&fee, "fee", "f", "", "transaction fee")
+	idCmd.Flags().Uint64Var(&Nonce, "nonce", 0, "nonce")
+}
+
+func generateIDAction() error {
+	pubkeyHex := pubkey
+	pubkey, err := hex.DecodeString(pubkeyHex)
```
gdmmx, line 58: A little confusion for readability.
`pubkey in L57` is a global var type string.
`pubkey in L58` is a local var type byte slice.
Local var override global var at here.

yilunzhang:
With the new structure, all commands (e.g. `nknc wallet`, `nknc asset`, etc) are defined in the same package, so all of them share private variables (e.g. `fee` for all commands are defined in `asset.go`). It's better to move those shared variable (e.g. fee, nonce) to a file that is not specific to a command, like `root.go`, this way removing or change one command won't affect the other ones.

omani:
> With the new structure, all commands (e.g. `nknc wallet`, `nknc asset`, etc) are defined in the same package, so all of them share private variables (e.g. `fee` for all commands are defined in `asset.go`). It's better to move those shared variable (e.g. fee, nonce) to a file that is not specific to a command, like `root.go`, this way removing or change one command won't affect the other ones.

yeah I already moved all shared vars to `root.go`. and kept all local (package scoped) vars to the appropriate files. to distinguish them from global vars.

so, as I understand it, you want me to make all vars global?

yilunzhang:
> yeah I already moved all shared vars to root.go. and kept all local (file scoped) vars to the appropriate files. to distinguish them from global vars. or in other words: if any var is being used in more than one file, it is declared in root.go. any other var is in the file itself.

This is exactly what I'm thinking as well, but there are a few variables that are used in more than one files but not defined in `root.go`. Specifically, `fee` and `nonce` are defined in `asset.go` but used in multiple files. I'm not sure if there are more variables like this but we can all check that.

omani:
> This is exactly what I'm thinking as well, but there are a few variables that are used in more than one files but not defined in `root.go`. Specifically, `fee` and `nonce` are defined in `asset.go` but used in multiple files. I'm not sure if there are more variables like this but we can all check that.

ok. I will check everything and commit soon.

omani:
> This is exactly what I'm thinking as well, but there are a few variables that are used in more than one files but not defined in `root.go`. Specifically, `fee` and `nonce` are defined in `asset.go` but used in multiple files. I'm not sure if there are more variables like this but we can all check that.

done

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	default:
+	}
+
+	FormatOutput(resp)
+
+	return nil
+}
+
+func getPassword(passwd string) []byte {
```
yilunzhang, line 198: The same function is duplicated in `root.go`

cmd/nknc/commands/root.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/howeyc/gopass"
+	"github.com/nknorg/nkn/v2/config"
+	"github.com/spf13/cobra"
+)
+
+// Globals
+var (
+	WalletFile     string
```
yilunzhang, line 20: I suggest we make all these variables private unless they are needed outside package (which seems not?)

omani: done.

cmd/nknc/commands/asset.go
omani: thanks.

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		if assetname == "" {
+			fmt.Println("asset name is required with [--name]")
+			return nil
+		}
+
+		if symbol == "" {
+			fmt.Println("asset symbol is required with [--symbol]")
+			return nil
```
yilunzhang, line 119: L119 returns nil while L125 returns err. Is this expected?

cmd/nknc/commands/asset.go
```diff
@@ -0,0 +1,195 @@
...
+			fmt.Fprintln(os.Stderr, err)
+			return err
+		}
+
+		resp, err = client.Call(Address(), "sendrawtransaction", 0, map[string]interface{}{"tx": hex.EncodeToString(buff)})
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err)
+			return err
+		}
+	default:
```
yilunzhang, line 189: If no sub command is specified, the previous behavior is to print usage for this command, while the new behavior seems to be print nothing?

cmd/nknc/commands/asset.go
omani: this is from the original code. I just took it as it is:
```
		if symbol == "" {
			fmt.Println("asset symbol is required with [--symbol]")
			return nil
		}

		totalSupply, err := common.StringToFixed64(value)
		if err != nil {
			fmt.Fprintln(os.Stderr, err)
			return err
		}
```

L125 has an actual error. L119 is just a check if flag is set. this makes sense to me.

cmd/nknc/commands/asset.go
omani: yes, this is by design. if you look at L27 you will see that we use `RunE` and check for subcommands. if none present, we print usage, by the help of `showusage` boolean:

```
RunE: func(cmd *cobra.Command, args []string) error {
		showusage := true
		cmd.Flags().Visit(func(name *pflag.Flag) {
			showusage = false
		})
		if showusage {
			return cmd.Usage()
		}
		if err := assetAction(); err != nil {
			log.Error(err)
		}
		return nil
	},
```

this applies to all commands.

cmd/nknc/commands/asset.go
yilunzhang: But the behavior of cobra seems to be different from the library used before. Is it going to show the help text if you return error here? We only want error to be printed, but not help text

cmd/nknc/commands/root.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/howeyc/gopass"
+	"github.com/nknorg/nkn/v2/config"
+	"github.com/spf13/cobra"
+)
+
+// Globals
+var (
+	walletFile     string
+	walletPassword string
+	Name           string
```
yilunzhang, line 22: Do we need these variables (L22-L26) to be public?

cmd/nknc/commands/asset.go
yilunzhang: To restore the previous behavior, we probably need to print asset cmd usage in this default branch. Same for all other commands default branch.

omani: > To restore the previous behavior, we probably need to print asset cmd usage in this default branch. Same for all other commands default branch.

yes. thats what I did. couldnt finish as I was busy lately. changes will come soon.

yilunzhang (approved):
Looks good to me

omani:
rebased

gdmmx (approved):
LGTM

yilunzhang:
Upon further review it seems that print version feature is missing. Previously `nknc --version` and `nknd --version` will print out nknc/nknd version (set in Makefile https://github.com/nknorg/nkn/blob/master/Makefile#L13), but I can't find it in the new code. Is it something missing here?

omani:
> Upon further review it seems that print version feature is missing. Previously `nknc --version` and `nknd --version` will print out nknc/nknd version (set in Makefile https://github.com/nknorg/nkn/blob/master/Makefile#L13), but I can't find it in the new code. Is it something missing here?

```
$ go run main.go --version
nknc version 1.0
```

```
$ go run main.go --version
nknd version 1.0
```

is what I get.

yilunzhang:
Yep that's what I meant by missing because it's always 1.0... The expected behavior is to print version set in Makefile, to be more specific, it should be git tag and commit hash. You should pass version when creating cobra.Command (to `config.Version`) and also change the part in Makefile L14 from
`github.com/nknorg/nkn/v2/cmd/nknc/common.Version=$(VERSION)`
to
`github.com/nknorg/nkn/v2/config.Version=$(VERSION)`

yilunzhang:
For now they are the same, but as you can see they might contain different build parameters in the future, so it's better to keep both `NKND_BUILD_PARAM` and `NKNC_BUILD_PARAM`. If you want to reduce redundancy I would suggest something like this:
```
NKNC_BUILD_PARAM=$NKND_BUILD_PARAM
```

yilunzhang:
Looks good to me!

PR #859, opened 2022-06-21
Add per ip rate limiter

yilunzhang:


node/localnode.go
```diff
@@ -110,6 +112,18 @@ func NewLocalNode(wallet *vault.Wallet, nn *nnet.NNet) (*LocalNode, error) {
 	event.Queue.Subscribe(event.BlockPersistCompleted, localNode.cleanupTransactions)
 	event.Queue.Subscribe(event.NewBlockProduced, localNode.CheckIDChange)
 
+	nn.MustApplyMiddleware(nnetnode.ConnectionAccepted{func(conn net.Conn) (bool, bool) {
+		host, _, err := net.SplitHostPort(conn.RemoteAddr().String())
```
bufrr, line 116: `host` including IP and port, maybe we need to limit the IP access rate without port

yilunzhang: host does not include port, check https://pkg.go.dev/net#SplitHostPort

api/websocket/server/server.go
```diff
@@ -247,6 +248,16 @@ func (ws *WsServer) Stop() {
 
 //websocketHandler
 func (ws *WsServer) websocketHandler(w http.ResponseWriter, r *http.Request) {
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err == nil {
```
bufrr, line 252: Do we need to check if `err != nil` ?

yilunzhang: In theory it shouldn't happen, and even it does, we can't do anything about it except for print it...

api/httpjson/RPCserver.go
```diff
@@ -59,7 +60,18 @@ func NewServer(localNode *node.LocalNode, wallet *vault.Wallet) *RPCServer {
...
+		limiter := ratelimiter.GetLimiter("rpc:"+host, config.Parameters.RPCIPRateLimit, int(config.Parameters.RPCIPRateBurst))
+		if !limiter.Allow() {
+			log.Infof("RPC connection limit of %s reached", host)
+			w.WriteHeader(http.StatusTooManyRequests)
+			return
+		}
+	}
+
 	if !s.limiter.Allow() {
+		log.Infof("RPC connection limit reached", host)
```
bufrr, line 74: `host` havn't added to the log string

yilunzhang: Fixed, no need to log host because this is overall rate limit

bufrr (approved):

api/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,24 @@
...
+
+	if limiter, ok := rateLimiters[key]; ok {
+		return limiter
+	}
+
+	limiter := rate.NewLimiter(rate.Limit(limit), burst)
+	rateLimiters[key] = limiter
+
+	return limiter
+}
```
gdmmx, line 24: Which timing to remove key/Limiter obj from Map even conn/request had finished?

gdmmx (approved):
LGTM

PR #397, opened 2018-12-19
Resolved two Issue #389 #396

gdmmx:
### Proposed changes in this pull request
Resolved Issue #389 
Resolved Issue #396

### Type (put an `x` where ever applicable)
- [x] Bug fix: Link to the issue
- [x] Feature (Non-breaking change)
- [ ] Feature (Breaking change)
- [ ] Documentation Improvement

### Checklist
Please put an `x` against the checkboxes. Write a small comment explaining if its `N/A` (not applicable)

- [x] Read the [CONTRIBUTION guidelines](https://github.com/nknorg/nkn#contributing).
- [x] All the tests are passing after the introduction of new changes.
- [ ] Added tests respective to the part of code I have written.
- [ ] Added proper documentation where ever applicable (in code and README.md).
- [x] Code has been written according to [NKN-Golang-Style-Guide](https://github.com/nknorg/nkn/wiki/NKN-Golang-Style-Guide)

### Extra information
Any extra information related to this pull request.

PR #866, opened 2022-10-29
Feature: Add SyncMode to JSON RPC and nknc

### LINKED ISSUE #825: Add SyncMode to JSON RPC and nknc
### Description of the issue
Today the node JSON RPC does not return the value of SyncMode. It will be good to add this.

### Issue-Type (put a `x` sign in the square brackets)
- [ ] bug report
- [ x] feature request
- [ ] Documentation improvement

### Checklist
- [ x] Running latest version of code.
- [ ] This issue has not been reported earlier.

### Your environment
* OS: Ubuntu Linux 18.04 LTS
* Go version: 
* Release tag/commit of the code: 2.1.7

omani:
Fixes #825 

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ cscope*
 /wallet.json
 /wallet.pswd
 /config.json
+/cmd/nknc/nknc
+/cmd/nknd/nknd
+test.json
--- a/api/common/interfaces.go
+++ b/api/common/interfaces.go
@@ -405,6 +405,13 @@ func setDebugInfo(s Serverer, params map[string]interface{}, ctx context.Context
 	return respPacking(errcode.SUCCESS, nil)
 }
 
+// getSyncMode gets sync mode of node
+// params: {}
+// return: {"resultOrData":<result>|<error data>, "error":<errcode>}
+func getSyncMode(s Serverer, params map[string]interface{}, ctx context.Context) map[string]interface{} {
+	return respPacking(errcode.SUCCESS, config.Parameters.SyncMode)
+}
+
 // getVersion gets version of this server
 // params: {}
 // return: {"resultOrData":<result>|<error data>, "error":<errcode>}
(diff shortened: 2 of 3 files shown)


yilunzhang:
I have two suggestions:

1. Returning `config.Parameters.SyncMode` is not always the "expected" result. The goal of this API is to distinguish whether a node has full ledger or light ledger (i.e. whether it has full block history), but `config.Parameters.SyncMode` is only effective when node sync from scratch. For example, if a user changes `SyncMode` in `config.json` to `light`, but the node already has a full block history in ChainDB (not a fresh sync), nothing will happen. In this case `config.Parameters.SyncMode` is `light` while the node actually has full block history. Similarly, if a node uses light sync to get a light ledger, then change `SyncMode` in `config.json` to `full`, `config.Parameters.SyncMode` will be `full` while node the node doesn't have a full block history.
I'm proposing the following logic:
If `chainstore.GetHeight() == 0`, returns `light` if  `config.Parameters.SyncMode == "light"`;
If `chainstore.GetHeight() > 0`, return `light` if `GetBlockByHeight(1)` success, i.e. node has block at height 1.
@bufrr Do you think there is a more elegant solution?
Also, we need to set this value when node starts to avoid db read per RPC request.

2. It's probably better to include this result in `getnodestate` RPC response rather than making a new RPC method

bufrr:
> 1. SyncMode

I suggest we return `full` only if `config.Parameters.SyncMode == "full"` && `chainstore.GetHeight() > 0` and get 3-5 random blocks which below current hight from node and verify with block header. That must be more accurate and prevent light nodes fake their syncmode to get more peers.

abebeos:
As you're discussing this PR, I assume that this feature is to be implemented/merged.

My understanding so far:

* "set this value when node starts to avoid db read per RPC request."

```go
# pseudo code

if (chainstore.GetHeight() == 0 && config.Parameters.SyncMode == "light")
  return "light"
if (chainstore.GetHeight() > 0 && GetBlockByHeight(1) )
  return "light"

if config.Parameters.SyncMode == "full" && chainstore.GetHeight() > 0 && VerifyRandomBlocks(5)
  return "full"
```

A simplification would be to check only for "full" node (to "prevent light nodes fake their syncmode to get more peers") :

```go
# pseudo code

if isFullNode()
  return "full"
else
  return "light"



```

yilunzhang:
The conditions need to be simplified a bit, and we don't really need to worry about "fake" response because we don't need to trust it in the first place. It's just a reference for convenience. I'm not sure if the author of this PR is still working on it though.

abebeos:
> we don't really need to worry about "fake" response because we don't need to trust it in the first place.

Nice, then I'd agree with you on this

> t's probably better to include this result in getnodestate RPC response rather than making a new RPC method

Example: https://github.com/nknorg/nkn/pull/397/commits/769f9e1df14b0517a8775a6069bc7fbdee00d3e1

I could possibly take a look if @omani is not active on this anymore.

omani:
> > we don't really need to worry about "fake" response because we don't need to trust it in the first place.
> 
> Nice, then I'd agree with you on this:
> 
> > It's probably better to include this result in getnodestate RPC response rather than making a new RPC method
> 
> Example: [769f9e1](https://github.com/nknorg/nkn/commit/769f9e1df14b0517a8775a6069bc7fbdee00d3e1)
> 
> I could possibly take a look if @omani is not active on this anymore.

hi @abebeos , yes go for it. Im not active on this PR.

abebeos:
@yilunzhang , this becomes actually very easy with the new requirements (and with protobuf), see #904. 

Please feel free to use #904 or to close it an process the issue as you prefer.

(this PR here can be closed)

Issue #827, opened 2021-11-05
Implement client message cache expiration and memory limitation


yilunzhang:
### Issue-Type (put a `x` sign in the square brackets)

- [ ] bug report
- [x] feature request
- [ ] Documentation improvement

### Description of the issue

When a node receives a message that should be delivered to client, the node will cache the message if all of the following conditions are met:

1. node is the last hop
2. client is not online currently
3. message `MaxHoldingSeconds > 0`

Currently the cache is not fully implemented yet, see details below.

### Actual behaviour

Currently cache does not have expiration, all messages will be cached indefinitely as long as `MaxHoldingSeconds > 0`.

Currently cache also doesn't have any RAM usage limit.

### Expected behaviour

Cache should have space and time limit:

- A message should be removed from cache if it has been cached for `MaxHoldingSeconds` seconds and still not delivered.
- Node should be able to configure max cache size. When total cached message size is more than max cache size, messages with oldest add time should be removed until total cached message size is no more than max cache size.
- All operation (add message to cache, remove message from cache due to expiration or memory limit) should have at most `O(log N) ` time complexity where `N` is the number of messages in cache
- Comprehensive test case should be included

abebeos:
This one looks closable, too.

PR #867, opened 2022-11-02
Feature: Implement client message cache expiration and memory limitation

omani:
Proposal for #827.

### Checklist

- [x] Added proper documentation where ever applicable (in code and README.md).

### Extra information
- tested against race conditions
- test cases follow after general approval/review.

### DIFF
--- a/api/websocket/messagebuffer/messagebuffer.go
+++ b/api/websocket/messagebuffer/messagebuffer.go
@@ -1,218 +1,135 @@
 package messagebuffer
 
 import (
+	"container/heap"
 	"fmt"
 	"sync"
 	"time"
 
-	"github.com/nknorg/nkn/v2/config"
 	"github.com/nknorg/nkn/v2/pb"
 )
 
-type Item struct {
-	relay        *pb.Relay
-	creationTime time.Time
-	expiryTime   time.Time
-}
-
-type queue struct {
-	parent *Buffer
-	item   *Item
-}
-
 type Buffer struct {
-	mux   sync.Mutex
-	queue map[string][]*queue
-	size  int
+	mux    sync.Mutex
+	equeue *ExpirationQueue
+	mqueue map[string]*MessageQueue
+	size   int
 }
 
 func New() *Buffer {
+	eq := make(ExpirationQueue, 0)
+	heap.Init(&eq)
+
 	return &Buffer{
-		queue: make(map[string][]*queue, 0),
+		equeue: &eq,
+		mqueue: make(map[string]*MessageQueue, 0),
 	}
 }
 
-func getParent(index int) int {
-	if index <= 0 {
-		return -1
-	}
+func (b *Buffer) AddMessage(clientID string, relay *pb.Relay) error {
+	b.mux.Lock()
+	defer b.mux.Unlock()
 
-	return (index - 1) / 2
-}
-
-func getChildren(index int) (int, int) {
-	tree := index * 2
-
-	return tree + 1, tree + 2
-}
-
-func (q *queue) isAfter(node *queue) bool {
-	a := q.item
-	b := node.item
-
-	if a.creationTime.Before(b.creationTime) {
-		return false
-	}
-	if a.creationTime == b.creationTime {
-		return false
-	}
-
-	return true
-}
-
-func (q *queue) isAfterOrEqual(node *queue) bool {
-	a := q.item
-	b := node.item
-
-	if a.creationTime.Before(b.creationTime) {
-		return false
-	}
-	if a.creationTime == b.creationTime {
-		return true
-	}
-
-	return true
-}
-
-func (m *Buffer) Purge() {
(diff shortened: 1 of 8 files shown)


api/websocket/messagebuffer/messagebuffer.go
```diff
@@ -1,39 +1,218 @@
...
+		return false
+	}
+	if a.creationTime == b.creationTime {
+		return true
+	}
+
+	return true
+}
+
+func (m *Buffer) Purge() {
```
yilunzhang, line 77: Purge has O(N) time complexity, which is too high as a regular cleaning job

omani: correct. I have my trouble with purging this way. I am thinking about a better way to solve this. I left `Purge()` in there to demonstrate/test that an iteration over the whole map is needed to keep a clean cache state.

another way would be an irregular O(n) by doing the cleanup periodically and sacrificing accuracy.
yet another way would be to purge only when cache is full.

any ideas are welcome.

api/websocket/messagebuffer/messagebuffer.go
```diff
@@ -1,39 +1,218 @@
...
-	clientIDStr := hex.EncodeToString(clientID)
-	messageBuffer.Lock()
-	defer messageBuffer.Unlock()
-	messageBuffer.buffer[clientIDStr] = append(messageBuffer.buffer[clientIDStr], msg)
+func (m *Buffer) AddMessage(clientID string, relay *pb.Relay) error {
+	if relay.MaxHoldingSeconds <= 0 {
+		return nil
+	}
+
+	m.Purge()
```
yilunzhang, line 100: Calling Purge in AddMessage have two unwanted results:
1. AddMessage has O(N) time complexity
2. If no new messages arrive, old messages will never be removed

api/websocket/messagebuffer/messagebuffer.go
```diff
@@ -1,39 +1,218 @@
...
 }
 
-// PopMessages reads and clears all messages of a client
-func (messageBuffer *MessageBuffer) PopMessages(clientID []byte) []*pb.Relay {
-	clientIDStr := hex.EncodeToString(clientID)
-	messageBuffer.Lock()
-	defer messageBuffer.Unlock()
-	messages := messageBuffer.buffer[clientIDStr]
-	messageBuffer.buffer[clientIDStr] = nil
+func (m *Buffer) PopMessages(clientID string) []*pb.Relay {
```
yilunzhang, line 109: You are trying to implement a heap right? Why not using the built in heap? https://pkg.go.dev/container/heap

api/websocket/messagebuffer/messagebuffer.go
```diff
@@ -1,39 +1,218 @@
...
+		},
+	}
+
+	m.queue[clientID] = append(m.queue[clientID], node)
+
+	par := getParent(next)
+	cur := next
+
+	for (par >= 0) && (q[par].isAfter(node)) {
+		m.queue[clientID][par], m.queue[clientID][cur] = q[cur], q[par]
```
yilunzhang, line 156: This seems not right. len(q) is not increased after append (but len(m.queue[clientID]) is), so you are accessing q[len(q)]

api/websocket/messagebuffer/messagebuffer.go
omani: yes. this is a heap implementation. I didnt think of container/heap.

api/websocket/messagebuffer/messagebuffer.go
yilunzhang: You can use another heap (sorted by expiration time) for fast (O(log N)) purge. The caveat is that when removing an element from one heap, you need to remove it from the other one as well.

omani:
switched to container/heap. commit not squashed to reflect PR discussion.

omani:
b.size <= 1000 is for testing.

omani:
we need to discuss the purging.

yilunzhang:
One way to purge efficiently is using another heap sorted by expiration time. This way each purge call has O(log N) time complexity per element being removed. This heap can be shared by all message recipients. The caveat is that when removing an element from one heap, you need to remove it from the other one as well.

omani:
> One way to purge efficiently is using another heap sorted by expiration time. This way each purge call has O(log N) time complexity per element being removed. This heap can be shared by all message recipients. The caveat is that when removing an element from one heap, you need to remove it from the other one as well.

as you can see I have a second heap. but it is just a minheap of clients. when I maxheap for expirationtime, dont we remove clients that a newer but have longer maxholdingseconds? instead, this second heap is just removing the oldest client.

yilunzhang:
Yeah as you mentioned, the second heap isn't really helping us achieve fast purge

yilunzhang:
Closing this PR as it's implemented in #876

PR #904, opened 2023-01-25
add SyncMode to NodeData RPC, re #825

abebeos:
This is to replace PR #866.

makes me a bit nervous, as this produces protobuf binary stuff, though should be non-critical.

(please feel free to close this unmerged and process this on your side as you prefer, done this mostly out of curiosity)


yilunzhang:
My comment in #866 also applies here: https://github.com/nknorg/nkn/pull/866#issuecomment-1296073969

So basically what we should include in response is node's ledger state, i.e. whether node has a full block history or not. This cannot be determined by the config. The only way to determine that is to try getting old blocks and see successful or not.

What I would suggest is that, we add a enum in node.proto

```
enum LedgerMode {
  full = 0;
  light = 1;
}
```

and change `sync_mode` to `ledger_mode` to avoid naming confusion.

When setting ledger mode, we use the following logic:

```
If chainstore.GetHeight() > 0
  if GetBlockByHeight(1) success
    return full
  else
    return light
If chainstore.GetHeight() == 0
  if config.Parameters.SyncMode == "light"
    return light
  else
    return full
```

Also we would need to make sure the json response returns human readable string rather than enum key, like https://github.com/nknorg/nkn/blob/master/node/node.go#L58

These changes requires some familiarity with existing code base, so feel free to let me know if you'd like me to help creating a new PR.

abebeos:
> and change `sync_mode` to `ledger_mode` to avoid naming confusion.

+1

> These changes requires some familiarity with existing code base, so feel free to let me know if you'd like me to help creating a new PR.

Please feel free to close this PR and implement `ledger_mode` the the way you see fit, I'm fine with it.

yilunzhang:
Implemented in new PR #905

PR #905, opened 2023-01-27
add ledger mode to node data and getnodestate rpc response

yilunzhang:
Implement #825


abebeos:
sync_mode_cfg
sync_mode

I guess when we say `node.sync_mode`, we refer to the real sync_mode of the node (what you call `ledger_mode`).

In order to distinguish from the configured sync_mode, it could be called `sync_mode_cfg`.

Adding both, `sync_mode_cfg` AND `sync_mode` (your `ledger_mode`) would give more flexibility, e.g. allowing to analyze via RPC if a node syncs in the configured mode.

```
if node.sync_mode != node.sync_mode_cfg
   Err: something is wrong
```

As always, please feel free to ignore fully.

Your PR looks very clean.

yilunzhang:
IMO there is no need to include sync_mode config in any of the RPC response at all because it's indistinguishable after node has synced once. Actually it causes a lot of confusion because use can change that config field after first sync but it won't change anything.

So basically sync mode is how node first acquire ledger, and ledger mode is what ledger node has. Once a node has ledger, it doesn't matter how he acquired the ledger initially, the only thing that matters is what ledger the node has.

pb/node.proto
```diff
@@ -10,6 +10,11 @@ enum SyncState {
   PERSIST_FINISHED = 3;
 }
 
+enum LedgerMode {
+  full = 0;
```
bufrr, line 14: Full synchronization contains two modes: full sync and fast sync, do we need to distinguish them?

yilunzhang: I don't think so. They are identical once sync finished

bufrr (approved):

Issue #919, opened 2023-02-06
Reduce Duplicate Code via Strict Refactoring


abebeos:
**Is your feature request related to a problem? Please describe.**

Code Duplication has several negativ

**Describe the solution you'd like**

Reduce code duplication.

* Use strict refactoring
  * Introduces zero functional changes
* ensure multiple commits that are easy to review

**Describe alternatives you've considered**

There is no real alternative.

**Additional context**

Existent core-developers usually focus on new functionality/enhancements/business-contracts (thus they usually don't go back into existent code for quality increase).

While the result looks many times like low effort, the nature of refactoring makes it necessary to do some dry runs, rollback fully and start again (with the knowledge gained) etc.

## Code Files

Well, I'm positively surprised. Re code-duplication, the nkn source-code-quality is high. Not much to do.

Listing the top-duplication files first:

### edwards25519.go

https://github.com/nknorg/nkn/blob/aa832aae9a4001c70484c9b3001de89fc3f31eb8/crypto/ed25519/edwards25519/edwards25519.go#L1-L2

~A few tests show that the file can be brought down to around 60% of its initial size, without significant performance loss (possibly even 0).~

The effort is a bit higher, due to the nature of the code (one needs to be a bit more careful, to have a fast and convenient result).

But this file is finally not about duplication. The quality problem is in the age of the file (2013), thus:

- [ ] #923

### Other Files

- [ ] https://github.com/nknorg/nkn/pull/922

PR #908, opened 2023-01-30
fix issue according to the analyzer

bufrr:

PR #921, opened 2023-02-06
remove duplicate code within 'chain/trie/pruning.go'

abebeos:
introduce closure function decrementRefCount for reuse this can be further refactored, though the effort can soon become too high

re #919 

(note that the generated diffs make it very difficult to review. It's better to place old/new side by side I guess.

## How to Not do it

This PR showcases how to **NOT do it**.

Though the interaction with the core-developer @yilunzhang leads to results, the effort is too high.

### Wrongs

* The PR is a `WIP` (Work In Progress), and thus changing during review
* The commits were squashed far too early
  * The diff is very difficult to review 

### Rights

* Closing this PR, and following up with a "fresh" one, with many commits, easier to review
* The new PR can later serve as a demo "how to do it right"

Follow-up PR: #922

### DIFF
--- a/chain/trie/pruning.go
+++ b/chain/trie/pruning.go
@@ -218,68 +218,60 @@ func (ref *RefCounts) Prune(hash common.Uint256, inMemory bool) error {
 	return nil
 }
 
+// decrement retrieves and decrements the ref.counts for the given node
+// returns false if counter reached zero, otherwise true
+func (ref *RefCounts) decrement(n node, inMemory bool) bool {
+	hash, _ := n.cache()
+	hs, _ := common.Uint256ParseFromBytes(hash)
+	count, ok := ref.counts[hs]
+	if !inMemory {
+		if !ok {
+			v, err := ref.trie.db.Get(db.TrieRefCountKey(hash))
+			if err == nil {
+				count = binary.LittleEndian.Uint32(v)
+				ref.counts[hs] = count
+			} else {
+				log.Fatal("Trie get error: %v", err)
+			}
+		}
+	}
+	if count == 0 {
+		log.Fatalf("RefCount cannot be zero, %v", hs.ToHexString())
+	}
+
+	if count > 1 {
+		ref.counts[hs] = count - 1
+		return true
+	}
+	return false
+}
+
+func (ref *RefCounts) delete(n node, inMemory bool) bool {
+	hash, _ := n.cache()
+	hs, _ := common.Uint256ParseFromBytes(hash)
+
+	delete(ref.counts, hs)
+	ref.trie.db.BatchDelete(db.TrieNodeKey(hash))
+	if !inMemory {
+		ref.trie.db.BatchDelete(db.TrieRefCountKey(hash))
+	}
+	return false
+}
+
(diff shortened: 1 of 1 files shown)


abebeos:
Ok, while self-reviewing, I realized that this could be better done by implementing a method `func (ref *RefCounts) dectrementRefCount bool`, which would have even higher reuse and would be a cleaner design:

https://github.com/nknorg/nkn/blob/5fec0648dbcaedfb6abc176a20c3d3263b17d8aa/chain/trie/pruning.go#L224-L246

abebeos:
question (curiosity) : would it be possible to keep `inMemory` within `RefCounts` ?

abebeos:
I detected similar code, and a recent change:

```go

	hash, _ := n.cache()
	hs, _ := common.Uint256ParseFromBytes(hash)

# vs

	hs, _ := common.Uint256ParseFromBytes(n)
```

@bufrr , you had recently applied a change here:

https://github.com/nknorg/nkn/pull/908/files#diff-9dd315e904047313d67365f93a7efac926e9152b74642ceb6d2b8d75e08d2b5a

can this be used in general at other places like e.g.: 

https://github.com/nknorg/nkn/blob/aa832aae9a4001c70484c9b3001de89fc3f31eb8/chain/trie/pruning.go#L221-L225

(note that I refactor abstractly, mechanically, without going deeper into code to analyze).

chain/trie/pruning.go
```diff
@@ -218,77 +218,69 @@ func (ref *RefCounts) Prune(hash common.Uint256, inMemory bool) error {
...
+	return false
+}
 
-		delete(ref.counts, hs)
-		ref.trie.db.BatchDelete(db.TrieNodeKey(hash))
-		if !inMemory {
-			ref.trie.db.BatchDelete(db.TrieRefCountKey(hash))
-		}
-		return ref.prune(n.Val, inMemory)
+func (ref *RefCounts) delete(n node, inMemory bool) bool {
```
yilunzhang, line 249: Seems we don't need the return value

yilunzhang:
> would it be possible to keep inMemory within RefCounts ?

Logically it's better to keep them separated because one can prune the same refcount with different mode, optimizing for time or space complexity.

> Shouldn't
> 
> func (n hashNode) cache() (hashNode, bool) {
> 	return nil, true
> }
> return a hash, instead of nil?

cache() returns the cached linked node to avoid another db hit. hashNode and valueNode don't have linked node.

abebeos:
> > would it be possible to keep inMemory within RefCounts ?
> 
> Logically it's better to keep them separated because one can prune the same refcount with different mode, optimizing for time or space complexity.

Todo: Reduce or document

abebeos:
> > Shouldn't
> > func (n hashNode) cache() (hashNode, bool) {
> > return nil, true
> > }
> > return a hash, instead of nil?
> 
> cache() returns the cached linked node to avoid another db hit. hashNode and valueNode don't have linked node.

Todo: Reduce or document

abebeos:
closing in favor of Follow-up PR: #922

PR #922, opened 2023-02-07
Rework 'chain/trie/pruning.go' to reduce duplication - chunk 1

abebeos:
(re #919)
(#921 showed how to not do it.)

This PR reworks 'chain/trie/pruning.go'. This is not a strict refactoring, because the functionality will change, though only slightly. The context is "reduce duplication".

But as `Refactoring` means `zero functional change`, this here is called `Rework`.

The commit history is kept to showcase additionally the wrong steps. Additionally, the commits can be reviewed one by one, which is usually easier.

## General Review Rules

(why at it, let's produce some rules)

* Make review comments in context (via the `commits`/`Files changed` ui on github)
* Let the executer decide how to proceed (instead of micromanaging the process)
* do not review whilst having super-high quality standards - except those are already applied to the overall code (or at least the code surrounding your review)
* Chunks of Work (not everything must go into one PR)

### DIFF
--- a/chain/trie/pruning.go
+++ b/chain/trie/pruning.go
@@ -283,11 +283,12 @@ func (ref *RefCounts) prune(n node, inMemory bool) error {
 
 		return nil
 	case hashNode:
-		hs, _ := common.Uint256ParseFromBytes(n)
+		hash := n
+		hs, _ := common.Uint256ParseFromBytes(hash)
 		count, ok := ref.counts[hs]
 		if !inMemory {
 			if !ok {
-				v, err := ref.trie.db.Get(db.TrieRefCountKey(n))
+				v, err := ref.trie.db.Get(db.TrieRefCountKey(hash))
 				if err == nil {
 					count = binary.LittleEndian.Uint32(v)
 					ref.counts[hs] = count
--- a/chain/trie/pruning.go
+++ b/chain/trie/pruning.go
@@ -219,8 +219,8 @@ func (ref *RefCounts) Prune(hash common.Uint256, inMemory bool) error {
 }
 
 func (ref *RefCounts) prune(n node, inMemory bool) error {
-
-	decrementRefCount := func() bool {
+	switch n := n.(type) {
+	case *shortNode:
 		hash, _ := n.cache()
 		hs, _ := common.Uint256ParseFromBytes(hash)
 		count, ok := ref.counts[hs]
@@ -241,7 +241,7 @@ func (ref *RefCounts) prune(n node, inMemory bool) error {
 
 		if count > 1 {
 			ref.counts[hs] = count - 1
-			return true
+			return nil
 		}
 
 		delete(ref.counts, hs)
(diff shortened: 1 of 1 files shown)


yilunzhang (commented):
Since this is considered as a rework, it's better to fix the issues in original code.

I wouldn't suggest to have `nodeRefHashHs` function at all. The meaning of returned hash/hs is not very clear and depend on n type. In the original code hash and hs are the same bytes but in different types (slice and array). In general it's better just keep and passing the array since array to slice conversion doesn't need to copy underlying buffer, so no buffer copy overhead. I would say we just get the hash array in original select case, pass it to decrement/delete function, and use `.ToArray()` in those functions to convert the hash array to hash bytes. This way we can get rid of the buffer copy overhead introduced in this PR.

Also we should handle the error returned by `Uint256ParseFromBytes`. Simply catch and return should be enough.

abebeos:
> Since this is considered as a rework, it's better to fix the issues in original code.

Thing is that this anything can be done in a followup PR. All this is an incremental process.

Main objective now is the duplication,

> I wouldn't suggest to have `nodeRefHashHs` function at all.
[...]

Review comments should be placed inline into context.

(I'd prefer to revert the last step for now, if you think the speed parameter is critical.)

abebeos:
Again general comment: Mention the problem, in context via review. Let the executer decide how to proceed (instead of micromanaging the process)

abebeos:
> The meaning of returned hash/hs is not very clear and depend on n type.

(again general rule) : do not review whilst having super-high quality standards - except those are already applied to the overall code.

here: the naming of "hash" and "hs" is unclear within the code. variables need to be renamed. But i cannot raise the quality of the file within one PR - this will take forever (ok, possibly week/month).

abebeos:
> Since this is considered as a rework, it's better to fix the issues in original code.

Possibly, but this interferes with my process. Your review comments need to support my processing, not introduce a new line of work. I do an architectural rework, so the lower details are currently out of scope (can be done later, even by a junior dev). So I'd need just a review re "problems introduced due to changes".

Re buffer-copy:
* `nodeRefHashHs` is most possibly inlined, so no buffer copy at all. But did not take the effort to verify/be sure (if it's inlined) , because...
  * ...it is an intermediate function (which suggests further design problems, necessary steps, renaming, etc.)

abebeos:
> In the original code hash and hs are the same bytes but in different types (slice and array)

The original code is actually quite confusing. `hash` vars have different types (same author uses both, `hs` and `hash` vars for Uint256. Anyways, tiny detail.

I assume I can introduce a new method to the `node` type? This would actually fix all my problems.

abebeos:
ok, fought/thought through some trials moving `refCountsIndex` to the type `node`, but this makes things actually just more complex.

There a some more duplications that can be solved via refactoring, but I'm close to a cognitional overload, so I need to abort and go for a merge of this PR.

chain/trie/pruning.go
```diff
@@ -218,68 +218,79 @@ func (ref *RefCounts) Prune(hash common.Uint256, inMemory bool) error {
...
-					log.Fatal("Trie get error: %v", err)
-				}
+	case *fullNode:
+		hash, _ = n.cache()
+	case hashNode:
+		hash = n
+	default:
+		log.Fatalf("Invalid node type : %v, %v", reflect.TypeOf(n), n)
+	}
+	hs, _ = common.Uint256ParseFromBytes(hash)
```
yilunzhang, line 235: Error is still not handled

yilunzhang (commented):
Since you created `refCountsIndex` function, you should use it for both refcount increment and decrement to make things consistent.

chain/trie/pruning.go
abebeos: And will not, I'm not a student of your's, stop this. https://github.com/nknorg/nkn/issues/924

abebeos:
> There a some more duplications that can be solved via refactoring, but I'm close to a cognitional overload, so I need to abort and go for a merge of this PR.

!!!

yilunzhang:
Code quality shouldn't depend on future changes that haven't been made yet

PR #930, opened 2023-02-23
unittest for election.go

bufrr:


yilunzhang (commented):
The test here doesn't test voting logic or results. Each node needs to send out votes to other nodes. You also need to test different initial votes and check if they converge to the majority.

consensus/election/election_test.go
```diff
@@ -0,0 +1,217 @@
...
+			result, _, _, _ := neighbors[i].GetResult()
+			if result != h {
+				t.Fatal("result is not equal to h")
+			}
+		}(i)
+	}
+	wg.Wait()
+}
+
+func newElection() (*Election, error) {
```
yilunzhang, line 153: Why not use NewElection?

consensus/election/election_test.go
```diff
@@ -0,0 +1,217 @@
...
+	vote2 := vote{
+		neighborID: "neighbor2",
+		hash:       common.MaxUint256,
+	}
+	vote3 := vote{
+		neighborID: "neighbor3",
+		hash:       h,
+	}
+
+	neighbors[0].txVoteChan <- vote1
```
yilunzhang, line 129: This is not right. You are not supposed to send stuff into the chan

consensus/election/election_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"reflect"
+	"sync"
+	"testing"
+	"time"
+)
+
+var neighbors []*Election
+
+func TestMain(m *testing.M) {
+	neighbors = make([]*Election, 3)
```
yilunzhang, line 18: Need to test on larger scale

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = elc.SetInitialVote(common.MaxUint256)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = elc.PrefillNeighborVotes([]interface{}{"100", "200", "300"}, common.EmptyUint256)
```
yilunzhang, line 42: It's better to use a non empty uint256 as vote

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+	ChangeVoteMinRelativeWeight: 0.5,
+	ConsensusMinRelativeWeight:  2.0 / 3.0,
+}
+
+func TestMain(m *testing.M) {
+	os.Exit(m.Run())
+}
+
+func TestSingleElection(t *testing.T) {
+	check := func(f string, got, want interface{}) {
```
yilunzhang, line 26: We can define this function outside to remove duplicate (L26 and L60)

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		var v common.Uint256
+		if i < 16 {
+			v = v1
+		} else {
+			v = common.MaxUint256
+		}
+		err = elcList[i].PrefillNeighborVotes([]interface{}{"100", "200", "300"}, v)
```
yilunzhang, line 87: This is not right. Each election should know the correct id and vote of all other nodes.

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		elcList[i].Start()
+	}
+
+	time.Sleep(1 * time.Second)
+
+	for i := 0; i < len(elcList); i++ {
```
yilunzhang, line 96: Each election should run this in a go routine to make it concurrent

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+		case vote := <-vchan:
+			for j, n := range elcList {
+				if i != j {
+					err = n.ReceiveVote(strconv.Itoa(i), vote)
+					if err != nil {
+						t.Fatal(err)
+					}
+				}
+			}
+		default:
```
yilunzhang, line 108: You shouldn't have this. vote chan will be closed when election is stopped.

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+		if i < 16 {
+			v = v1
+		} else {
+			v = common.MaxUint256
+		}
+		err = elcList[i].PrefillNeighborVotes([]interface{}{"100", "200", "300"}, v)
+		if err != nil {
+			t.Fatal(err)
+		}
+		elcList[i].Start()
```
yilunzhang, line 91: Should check return value

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	time.Sleep(1 * time.Second)
```
yilunzhang, line 94: Shouldn't need this

consensus/election/election_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/nknorg/nkn/v2/config"
+	"os"
+	"reflect"
+	"strconv"
+	"testing"
+	"time"
+)
+
+var c = &Config{
+	Duration:                    config.ConsensusDuration / 2,
```
yilunzhang, line 14: You don't need 10s for in memory test. I would say we can use 5s.

consensus/election/election_test.go
```diff
@@ -0,0 +1,145 @@
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			txVoteChan := elcList[i].GetTxVoteChan()
+			for vote := range txVoteChan {
+				for j, n := range elcList {
+					if i != j {
+						err = n.ReceiveVote(strconv.Itoa(i), vote)
+						if err != nil {
+							t.Error(err)
```
yilunzhang, line 125: Fatal

consensus/election/election_test.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		success := elcList[i].Start()
+		if !success {
+			t.Fatal("failed to start election:", i)
+		}
+	}
+
+	var wg sync.WaitGroup
+	for i := 0; i < len(elcList); i++ {
+		i := i
```
yilunzhang, line 115: This is correct too but we can follow the same coding style in this repo and use function argument

consensus/election/election_test.go
```diff
@@ -0,0 +1,145 @@
...
+	wg.Wait()
+
+	for _, n := range elcList {
+		result, absWeight, relWeight, err := n.GetResult()
+		if err != nil {
+			t.Fatal(err)
+		}
+		check(t, "result", result, v1)
+		check(t, "absWeight", absWeight, uint32(20))
+		check(t, "relWeight", relWeight, float32(1))
```
yilunzhang, line 143: here too

consensus/election/election_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	result, absWeight, relWeight, err := elc.GetResult()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	check(t, "result", result, v1)
+	check(t, "absWeight", absWeight, uint32(4))
+	tolerance := 1e-9
+	check(t, "relWeight", math.Abs(float64(relWeight-float32(1))) < tolerance, true)
```
yilunzhang, line 74: I think we can make a new function like almostEqual

yilunzhang (approved):

PR #939, opened 2023-06-06
Add websocket session connectTime to avoid deleting cached message too early

billfort:
1. Add websocket session a member `connectTime` ;
2. Set connectTime when session `setClient` ;
3. When pop `messageDeliveredCache` message, check session `connectTime` to avoid deleting the message too early.

### DIFF
--- a/api/websocket/server/relay.go
+++ b/api/websocket/server/relay.go
@@ -105,16 +105,18 @@ func (ws *WsServer) sendInboundMessage(clientID string, inboundMsg *pb.InboundMe
 	return success
 }
 
-func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay) {
+func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay, shouldSign bool) {
 	clientID := relayMessage.DestId
 	msg := &pb.InboundMessage{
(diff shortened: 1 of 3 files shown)


api/websocket/server/relay.go
```diff
@@ -105,16 +105,18 @@ func (ws *WsServer) sendInboundMessage(clientID string, inboundMsg *pb.InboundMe
 	return success
 }
 
-func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay) {
+func (ws *WsServer) sendInboundRelayMessage(relayMessage *pb.Relay, checkShouldSign bool) {
```
yilunzhang, line 108: The name is a bit confusing. I would recommend using `shouldSign` (same as local variable)

api/websocket/server/relay.go
```diff
@@ -149,14 +151,15 @@ func (ws *WsServer) startCheckingLostMessages() {
 				threshold := time.Now().Add(-pongTimeout)
 				success := false
 				for _, client := range clients {
-					if client.GetLastReadTime().After(threshold) {
+					if client.GetLastReadTime().After(threshold) && client.ConnectTime.Before(threshold) {
```
yilunzhang, line 154: Should write a read lock protected getter function `client.GetConnectTime()`

yilunzhang (approved):
LGTM

PR #944, opened 2023-08-01
Provide UnmarshalJSON() for common.uint160

omani:
fixes error `json: cannot unmarshal string into Go struct field Account.ProgramHash of type common.Uint160` in case `nkn.Account` struct is to be unmarshalled.

### DIFF
--- a/common/uint160.go
+++ b/common/uint160.go
@@ -89,9 +89,21 @@ func IsValidHexAddr(s []byte) bool {
 	return false
 }
 
-func (f Uint160) MarshalJSON() ([]byte, error) {
+func (f *Uint160) MarshalJSON() ([]byte, error) {
 	str, err := f.ToAddress()
 	return []byte("\"" + str + "\""), err
+}
+
+func (f *Uint160) UnmarshalJSON(in []byte) (err error) {
+	if len(in) < 2 {
+		return errors.New("[Common]: Uint160UnmarshalJSON err, len < 2")
+	}
+	temp, err := ToScriptHash(string(in[1 : len(in)-1]))
+	if err != nil {
+		return err
+	}
+	f.SetBytes(temp.ToArray())
+	return nil
 }
 
 func (f *Uint160) ToAddress() (string, error) {
--- a/common/uint160_test.go
+++ b/common/uint160_test.go
@@ -1,0 +1,72 @@
+package common
+
+import (
+	"encoding/hex"
+	"testing"
+)
+
+// NKN address: NKNPkGps7i6yye6W2qSVsQUgiyqxWHEoTRby
+// ToHexString(): 867ff0ca31905faec269949de37d27773ba7394e
+const NKNADDRESS = "\"NKNPkGps7i6yye6W2qSVsQUgiyqxWHEoTRby\""
+const NKNADDRESS_HEX = "867ff0ca31905faec269949de37d27773ba7394e"
+
+func TestMarshalJSON(t *testing.T) {
+	// Create instance of Uint160
+	f := new(Uint160)
+
+	dat, err := hex.DecodeString(NKNADDRESS_HEX)
+	if err != nil {
+		t.Fatal(err)
+	}
+	f.SetBytes(dat)
+
+	// Expected value after Marshalling
+	expected := []byte(NKNADDRESS)
+
+	bytes, err := f.MarshalJSON()
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	if !bytesEqual(bytes, expected) {
+		t.Fatalf("Expected %v, got %v", expected, bytes)
+	}
+
+	// TODO: If possible, create a scenario where ToAddress returns an error
+	// and verify that MarshalJSON also returns this error.
+}
+
+func TestUnmarshalJSON(t *testing.T) {
+	// Input for unmarshalling
+	input := []byte(NKNADDRESS)
+
+	f := new(Uint160)
+	err := f.UnmarshalJSON(input)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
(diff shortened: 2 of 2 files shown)


common/uint160.go
```diff
@@ -94,6 +94,11 @@ func (f Uint160) MarshalJSON() ([]byte, error) {
 	return []byte("\"" + str + "\""), err
 }
 
+func (f Uint160) UnmarshalJSON(in []byte) (err error) {
+	f, err = ToScriptHash(string(in[1 : len(in)-1]))
```
yilunzhang, line 98: Need to validate `len(in) > 1` to prevent panic

omani: silly me. fixed.

common/uint160.go
```diff
@@ -94,6 +94,13 @@ func (f Uint160) MarshalJSON() ([]byte, error) {
 	return []byte("\"" + str + "\""), err
 }
 
+func (f Uint160) UnmarshalJSON(in []byte) (err error) {
+	if len(in) > 1 {
```
yilunzhang, line 98: We should return error if `len < 2`, otherwise user will see `err == nil` and thought unmarshal is successful...

omani: > We should return error if `len < 2`, otherwise user will see `err == nil` and thought unmarshal is successful...

I thought about that too but my way of thinking was that `UnmarshalJSON()` returns an error on an operation that fails (in your implementation of it) rather than signalling that there was no operation at all. meaning, if there is nothing to unmarshal, it shouldnt be an error.

your suggestion indicates that we should fail if there is nothing to unmarshal. eg. what if user intentionally `omitempty`s on his side?

omani: from `encoding/json`:
```
// Unmarshaler is the interface implemented by types
// that can unmarshal a JSON description of themselves.
// The input can be assumed to be a valid encoding of
// a JSON value. UnmarshalJSON must copy the JSON data
// if it wishes to retain the data after returning.
//
// By convention, to approximate the behavior of Unmarshal itself,
// Unmarshalers implement UnmarshalJSON([]byte("null")) as a no-op.
```

futhermore, from https://pkg.go.dev/encoding/json#Unmarshal:
```
To unmarshal JSON into a value implementing the Unmarshaler interface,
Unmarshal calls that value's UnmarshalJSON method, including when the input is a JSON null.
```

omani: my omitempty assumption was wrong. `omitempty` leaves out the field completely if empty. so it wouldnt be in the JSON anyway and thus wont attempt to unmarshal.

omani: done.

common/uint160.go
```diff
@@ -94,6 +94,14 @@ func (f Uint160) MarshalJSON() ([]byte, error) {
 	return []byte("\"" + str + "\""), err
 }
 
+func (f Uint160) UnmarshalJSON(in []byte) (err error) {
+	if len(in) < 2 {
+		return errors.New("[Common]: Uint160UnmarshalJSON err, len < 2")
+	}
+	f, err = ToScriptHash(string(in[1 : len(in)-1]))
```
yilunzhang, line 101: This will not update f. You need to change
```func (f Uint160)```
to
```func (f *Uint160)```
then change `f, err = ` to `temp, err :=`,
and finally call `f.SetBytes(temp.ToArray())` to update f.

**Comment by yilunzhang** [comment]
FYI I asked ChatGPT to write a unit test for this. This is the prompt I used:
```
Write unit tests for the following json marshal/unmarshal implementations:

func (f Uint160) MarshalJSON() ([]byte, error) {
	str, err := f.ToAddress()
 	return []byte("\"" + str + "\""), err
 }

 func (f Uint160) UnmarshalJSON(in []byte) (err error) {
 	if len(in) < 2 {
 		return errors.New("[Common]: Uint160UnmarshalJSON err, len < 2")
 	}
 	f, err = ToScriptHash(string(in[1 : len(in)-1]))
 	return
 }
```

Results I got:
```
package yourpackage

import (
	"testing"
	"errors"
)

func TestMarshalJSON(t *testing.T) {
	// Create instance of Uint160
	f := new(Uint160)
	
	// Expected value after Marshalling
	expected := []byte("\"expectedValue\"")

	bytes, err := f.MarshalJSON()
	if err != nil {
		t.Fatalf("Unexpected error: %v", err)
	}

	if !bytesEqual(bytes, expected) {
		t.Fatalf("Expected %v, got %v", expected, bytes)
	}

	// TODO: If possible, create a scenario where ToAddress returns an error
	// and verify that MarshalJSON also returns this error.
}

func TestUnmarshalJSON(t *testing.T) {
	// Input for unmarshalling
	input := []byte("\"validInput\"")

	f := new(Uint160)
	err := f.UnmarshalJSON(input)
	if err != nil {
		t.Fatalf("Unexpected error: %v", err)
	}

	// TODO: Verify that f has been updated correctly

	input = []byte("\"a\"")
	err = f.UnmarshalJSON(input)
	if err == nil || err.Error() != "[Common]: Uint160UnmarshalJSON err, len < 2" {
		t.Fatalf("Expected error, got: %v", err)
	}

	// TODO: If possible, create a scenario where ToScriptHash returns an error
	// and verify that UnmarshalJSON also returns this error.
}

func bytesEqual(a, b []byte) bool {
	if len(a) != len(b) {
		return false
	}
	for i, v := range a {
		if v != b[i] {
			return false
		}
	}
	return true
}
```

yilunzhang:
> I dont know how to do this since `SetBytes()` prevents us from ever having a wrong address. it will get filled with zeros and a malformed version will always result in address `NKNBV71h5oUBjknMHPikPU7nraw9KtksWjtB` with no error.

Don't worry, the comments are auto generated by ChatGPT too, not all of them apply to our case.

yilunzhang (approved):
Looks good to me

PR #966, opened 2024-05-09
Add OnOfferConnected channel for synchronize connected event for client

billfort:


api/webrtc/webrtc.go
```diff
@@ -99,8 +100,11 @@ func (c *Peer) Offer(label string) error {
 	if err != nil {
 		return err
 	}
+
+	c.OnOfferConnected = make(chan struct{}, 1)
```
yilunzhang, line 104: need to move this to constructer because very often people use it this way:
```
peer := NewPeer(...)
go func() { ... <-peer.OnOfferConnected ... }
peer.Offer()
```
so OnOfferConnected needs to be created when creating peer

billfort: Yes, modified.

api/webrtc/webrtc.go
```diff
@@ -99,8 +101,10 @@ func (c *Peer) Offer(label string) error {
 	if err != nil {
 		return err
 	}
+
 	dc.OnOpen(func() {
 		log.Debugf("Data channel %v has been opened\n", dc.Label())
+		c.OnOfferConnected <- struct{}{}
```
yilunzhang, line 107: need to use default to prevent blocking since Offer() can be called multiple times

billfort: Yes, modified.

yilunzhang (approved):

PR #995, opened 2025-04-11
refactor: use slices.Equal to simplify code

tsinghuacoder:
There is a [new function](https://pkg.go.dev/slices@go1.21.1#Equal) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/common/uint160_test.go
+++ b/common/uint160_test.go
@@ -3,6 +3,7 @@ package common
 import (
 	"encoding/hex"
 	"encoding/json"
+	"slices"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


yilunzhang:
Thanks for the contribution!

As you can see in https://github.com/nknorg/nkn/blob/master/go.mod, nkn requires go 1.20 to build, but slices was introduced in go 1.21.

But moving to go 1.21 is just a matter of time, so maybe we can keep this PR open for now, and merge it when we move to go 1.21.

tsinghuacoder:
> Thanks for the contribution!
> 
> As you can see in https://github.com/nknorg/nkn/blob/master/go.mod, nkn requires go 1.20 to build, but slices was introduced in go 1.21.
> 
> But moving to go 1.21 is just a matter of time, so maybe we can keep this PR open for now, and merge it when we move to go 1.21.

Thank you for your reply. The currently stable version of Go is 1.24. If needed, I would be happy to open a new PR to upgrade it.

